PR #224, opened 2022-04-22
define tls.Config explicitly instead redefine default TLS config options 

asmyasnikov:

The SDK's default tls.Config contains root CA. User may appends custom certificates to defaults TLS config, redefine min TLS version and change InsecureSkipVerify flag. 
Also internal config field secure is an ambigious with TLS config. Always true secure flag indirectly redefines tls config with default TLS, false secure flag skips TLS config to nil.
PR's contains changes for using user-defined tls.Config explicitly and deprecates other options about TLS config.

CHANGELOG:
* Added `WithTLSConfig` option for redefine TLS config
* Marked as deprecated options `WithSecure`, `WithInSecure`, `WithMinTLSVersion`, `WithTLSSInsecureSkipVerify`, `WithCertificate`, `WithCertificatesFromFile, `WithCertificatesFromPem`
* Added `sugar.LoadCertificatesFromFile` and `sugar.LoadCertificatesFromPem` helpers

## Pull request type

- [ ] Bugfix
- [x] Feature
- [x] Build related changes

Issue Number: #217 

## What is the new behavior?

-
-
-

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -24,20 +24,21 @@ import (
 type Config struct {
 	config.Common
 
-	trace                            trace.Driver
-	dialTimeout                      time.Duration
-	connectionTTL                    time.Duration
-	balancer                         balancer.Balancer
-	secure                           bool
-	endpoint                         string
-	database                         string
-	requestsType                     string
-	userAgent                        string
+	trace         trace.Driver
+	dialTimeout   time.Duration
+	connectionTTL time.Duration
+	balancer      balancer.Balancer
+	secure        bool
+	endpoint      string
+	database      string
+	requestsType  string
+	userAgent     string
+	grpcOptions   []grpc.DialOption
+	credentials   credentials.Credentials
+	tlsConfig     *tls.Config
+	meta          meta.Meta
+
 	excludeGRPCCodesForPessimization []grpcCodes.Code
-	grpcOptions                      []grpc.DialOption
-	credentials                      credentials.Credentials
-	tlsConfig                        *tls.Config
-	meta                             meta.Meta
 }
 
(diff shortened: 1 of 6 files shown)


config/config.go
```diff
@@ -218,12 +244,18 @@ func WithRequestsType(requestsType string) Option {
 	}
 }
 
+// WithMinTLSVersion applies minimum TLS version that is acceptable.
+//
+// Deprecated: use explicit WithTLSConfig option instead
 func WithMinTLSVersion(minVersion uint16) Option {
 	return func(c *Config) {
 		c.tlsConfig.MinVersion = minVersion
```
rekby, line 252: Here and in similar places - may be check c.tlsConfig != nil or force WithSecure(true) for prevent nil panic on config apply?

sugar/certificates.go
```diff
@@ -0,0 +1,50 @@
...
+	"encoding/pem"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/ydb-platform/ydb-go-sdk/v3/internal/xerrors"
+)
+
+func LoadCertificatesFromFile(caFile string) ([]*x509.Certificate, error) {
+	if len(caFile) > 0 && caFile[0] == '~' {
```
rekby, line 14: In common libraries doesn't need shell expand.
I suggent to use caFile as path "as is"

config/config.go
asmyasnikov: You right, ack

sugar/certificates.go
asmyasnikov: ack

PR #257, opened 2022-05-16
balanced code has been simplified

rekby:

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Custom tree of balancers. Very flexible, but unneeded system.

## What is the new behavior?
All code about select connection moved to router package.


internal/balancer/balancer.go
```diff
@@ -6,18 +6,17 @@ import (
...
-// Balancer is an interface that implements particular load-balancing
-// algorithm.
-//
-// Balancer methods called synchronized. That is, implementations must not
-// provide additional goroutine safety.
-type Balancer interface {
-	// Next returns next connection for request.
-	Next(ctx context.Context, opts ...NextOption) conn.Conn
+type (
+	Balancer = *Config
```
rekby, line 10: 1. It is internal type - nobody outside backage can declare var with the type - then it is safe to change interface to struct
2. Interface can be nil, some code use nil balancer - it need pointer to struct
3. No rename Balancer in the commit - for smaller code changes. It will rename in next commit (with rename/move type changes only).

internal/router/connections_state.go
```diff
@@ -0,0 +1,158 @@
...
+		return c, failedCount
+	}
+
+	c, _ := s.selectRandomConnection(s.lastAttempt, true)
+	return c, failedCount
+}
+
+func (s *connectionsState) preferConnection(ctx context.Context) conn.Conn {
+	if e, hasPreferEndpoint := ContextEndpoint(ctx); hasPreferEndpoint {
+		c := s.connByNodeID[e.NodeID()]
```
asmyasnikov, line 69: c, ok := s.connByNodeID[e.NodeID()]

internal/router/router.go
```diff
@@ -177,70 +168,81 @@ func (r *router) Invoke(
...
-			if cc.GetState() == conn.Banned {
-				r.cluster().Allow(ctx, cc)
-			}
-		} else {
-			if xerrors.MustPessimizeEndpoint(err, r.config.ExcludeGRPCCodesForPessimization()...) {
-				r.cluster().Ban(ctx, cc, err)
-			}
-		}
-	}()
+	return r.connectionsState
```
asmyasnikov, line 226: race condition - pointer may be used from outside of locks

asmyasnikov: I'm understand - connectionsState is immutable

asmyasnikov (approved):

PR #410, opened 2022-10-04
[testutil] add Table & Topic options comparator for testing

shmel1k:

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-


asmyasnikov (dismissed):

testutil/compare.go
```diff
@@ -294,3 +299,95 @@ func compareUUID(l, r *Ydb.Value) int {
...
 }
+
+// CompareCreateTableOptions compares table options used for creating YDB table.
+//
+// This function should be used only for testing purpose.
+//
+// # Experimental
+//
+// Notice: This API is EXPERIMENTAL and may be changed or removed in a later release.
+func CompareCreateTableOptions(expected, actual []options.CreateTableOption) func(*testing.T) {
```
rekby, line 310: I suggest return result: true/false + diff string.

1. What if I want check about options are not equal?
2. What if I want use it without T (for example internal tests with many times use testing.TB instead of *T), may be other test frameworks with use own types.

shmel1k: Decided to use something like

```go
type CreateTopicOptionsDesc struct {
	opts *rawtopic.CreateTopicRequest
}

func ToCreateTopicOptionsDesc(opts ...topicoptions.CreateOption) *CreateTopicOptionsDesc {
	req := &rawtopic.CreateTopicRequest{}
	for _, v := range opts {
		v(req)
	}

	return &CreateTopicOptionsDesc{
		opts: req,
	}
}
```

shmel1k:
@asmyasnikov , probably, we can try to fix this API. It does not look like experimental and should not be changed in future

testutil/controlutil/controlutil.go
```diff
@@ -0,0 +1,127 @@
...
+// ToCreateTableOptionsDesc creates new table options description used for creating YDB table.
+//
+// This function should be used only for testing purpose.
+//
+// # Experimental
+//
+// Notice: This API is EXPERIMENTAL and may be changed or removed in a later release.
+func ToCreateTableOptionsDesc(opts ...options.CreateTableOption) *CreateTableOptionsDesc {
+	req := &options.CreateTableDesc{}
+	all := &allocator.Allocator{}
```
asmyasnikov, line 35: allocator через New кажется надо получать
		a := allocator.New()
		defer a.Free()
чтобы пулы работали

shmel1k: fixed

PR #428, opened 2022-10-14
272 deduplicate ca loading

xopoww:

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: #272

## What is the new behavior?

- add `parseCertificateCached` to cache `*x509.Certificate`s by their SHA224
- add `readFileCached` to cache file contents by filenames
- add benchmarks to showcase CPU/memory improvements
- add test to check that certificate cache is safe for concurrent use

## Other information

Local run of the benchmark for certificate cache:
```
$ go test -benchmem -run=^$ -bench ^BenchmarkWithCertificateCache$ github.com/ydb-platform/ydb-go-sdk/v3
goos: linux
goarch: amd64
pkg: github.com/ydb-platform/ydb-go-sdk/v3
cpu: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
BenchmarkWithCertificateCache/no_cache-8                    1974            626424 ns/op           75745 B/op        602 allocs/op
BenchmarkWithCertificateCache/cached-8                      2080            596607 ns/op           70117 B/op        526 allocs/op
PASS
ok      github.com/ydb-platform/ydb-go-sdk/v3   2.690s
```
_(fluctuation of time/op results of this benchmark seems to be greater than the difference between test cases when run locally)_

Local run of the benchmark for file cache:
```
$ go test -benchmem -run=^$ -bench ^BenchmarkWithFileCache$ github.com/ydb-platform/ydb-go-sdk/v3
goos: linux
goarch: amd64
pkg: github.com/ydb-platform/ydb-go-sdk/v3
cpu: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
BenchmarkWithFileCache/no_cache-8                   1953            632330 ns/op           77944 B/op        613 allocs/op
BenchmarkWithFileCache/cert_cache-8                 2058            611228 ns/op           72311 B/op        537 allocs/op
BenchmarkWithFileCache/both_caches-8                1789            580403 ns/op           70056 B/op        527 allocs/op
PASS
ok      github.com/ydb-platform/ydb-go-sdk/v3   3.906s
```

Local run of concurrency test (with race detector):
```
$ go test -race -timeout 30s -run ^TestWithCacheConcurrent$ github.com/ydb-platform/ydb-go-sdk/v3
ok      github.com/ydb-platform/ydb-go-sdk/v3   0.111s
```


cached.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		return nil, xerrors.WithStackTrace(err)
+	}
+	if !DisableFileCache {
+		fileCache.Store(file, bytes)
+	}
+	return bytes, nil
+}
+
+var DisableCertificateCache = false
```
asmyasnikov, line 41: The same issue

connection_e2e_test.go
```diff
@@ -314,3 +316,157 @@ func TestConnection(t *testing.T) {
...
+				// child conns are closed on db.Close()
+				_, err := db.With(
+					ctx,
+					ydb.WithAnonymousCredentials(),
+					ydb.WithBalancer(
+						balancers.PreferLocationsWithFallback(
+							balancers.RandomChoice(), "a", "b",
+						),
+					),
+					ydb.WithSessionPoolSizeLimit(100),
```
asmyasnikov, line 364: Only this option enough for makes true child connection

connection_e2e_test.go
```diff
@@ -314,3 +316,157 @@ func TestConnection(t *testing.T) {
...
+		disableCertificateCache bool
+		disableFileCache        bool
+	}{
+		{"no cache", true, true},
+		{"cert cache", false, true},
+		{"both caches", false, false},
+	}
+	for _, tc := range tcs {
+		b.Run(tc.name, func(b *testing.B) {
+			ydb.DisableCertificateCache = tc.disableCertificateCache
```
asmyasnikov, line 404: You can write a private option such as withOnClose for tests different cases

asmyasnikov: For improved testing you can count cache hits over some callbacks.

cached.go
```diff
@@ -0,0 +1,69 @@
...
+	"crypto/sha256"
+	"crypto/x509"
+	"os"
+	"path/filepath"
+	"sync"
+
+	"github.com/ydb-platform/ydb-go-sdk/v3/internal/xerrors"
+)
+
+var DisableFileCache = false
```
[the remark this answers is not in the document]
asmyasnikov: The second problem with global variables - data race if user is a lot of silly. Because SDK's API allowed it. No dangerous API - no problems

cached.go
```diff
@@ -0,0 +1,69 @@
...
+
+var certificateCache sync.Map
+
+// parseCertificateCached is a cached version of x509.ParseCertificate. Cache key is
+//  sha256.Sum224(der)
+func parseCertificateCached(der []byte) (*x509.Certificate, error) {
+	// using Sum224 for cert data similarly to tls package
+	var sum [sha256.Size224]byte
+	if !DisableCertificateCache {
+		sum = sha256.Sum224(der)
```
asmyasnikov, line 51: What do you think about key as string(der)?
Any crypto hashing algorithms:
- uses a lot of CPU
- have collisions

asmyasnikov: May be search by looooooong string have the same CPU usage...

xopoww: So, a quick [bench](https://gist.github.com/xopoww/30f2ea0d88580c2d1323dfa392fd0424) shows that using `string(der)` is faster, but memory usage in this case is worse. Example bench result:
```
BenchmarkKeys/0/sha224/store____
BenchmarkKeys/0/sha224/store____-8                181789              5886 ns/op             392 B/op          8 allocs/op
BenchmarkKeys/0/sha224/load_hit_
BenchmarkKeys/0/sha224/load_hit_-8                203478              5412 ns/op              32 B/op          1 allocs/op
BenchmarkKeys/0/sha224/load_miss
BenchmarkKeys/0/sha224/load_miss-8                198758              5438 ns/op              32 B/op          1 allocs/op
BenchmarkKeys/0/string
BenchmarkKeys/0/string/store____
BenchmarkKeys/0/string/store____-8               1245193              1252 ns/op            2424 B/op          9 allocs/op
BenchmarkKeys/0/string/load_hit_
BenchmarkKeys/0/string/load_hit_-8               2395186               494.1 ns/op          2064 B/op          2 allocs/op
BenchmarkKeys/0/string/load_miss
BenchmarkKeys/0/string/load_miss-8               2732768               450.2 ns/op          2064 B/op          2 allocs/op
```

Another idea might be to somehow exploit binary format of x509 certificate without actually parsing it, using some sub-slice of data (which is guaranteed to be unique between certs) as key. But it seems to unstable to use.

cached.go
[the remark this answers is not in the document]
xopoww: Moved to internal and made most fileds/funcs unexported to also improve usage safety inside ydb.
The solution might be overengineered now, need feedback

connection_e2e_test.go
```diff
@@ -314,3 +316,145 @@ func TestConnection(t *testing.T) {
...
+			if err != nil {
+				t.Error(err)
+			}
+			wg.Done()
+		}()
+	}
+	wg.Wait()
+
+	want, got := nRoutines, cached.FileCache().Hits()
+	if want != got {
```
xopoww, line 457: This test is not good enough: it [fails](https://github.com/ydb-platform/ydb-go-sdk/actions/runs/3255017774/jobs/5343923180) on reconnect because certs are loaded again. Maybe should change to got >= want because we really only care about cache misses on child connections.

internal/certificates/certificates.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	value, exists := fileCache.Load(file)
+	if FileCacheHook != nil {
+		FileCacheHook(exists)
+	}
+	if exists {
+		certs, ok := value.([]*x509.Certificate)
+		if !ok {
+			panic("unknown file cache type")
```
rekby, line 39: Add prefix "ydb: " to panic text

internal/certificates/certificates.go
```diff
@@ -0,0 +1,126 @@
...
+	"os"
+	"sync"
+
+	"github.com/ydb-platform/ydb-go-sdk/v3/internal/xerrors"
+)
+
+var (
+	// fileCache
+	//  map[string][]*x509.Certificate
+	fileCache sync.Map
```
rekby, line 15: Certificates never removed from cache, it can cause memory leak in long running applications.

I suggest to remove certificates from cache by timer/item count.

certificates_test.go
```diff
@@ -0,0 +1,186 @@
+package ydb
```
rekby, line 1: The certificates_test.go is located far from certificates.go
What about move certificates_test.go to internal/certificates?

rekby (changes requested):

rekby:
Hello @xopoww, thanks for your work.

I suggest few changes before merge it.

certificates_test.go
```diff
@@ -0,0 +1,186 @@
...
+	tcs := []struct {
+		name    string
+		enabled bool
+	}{
+		{"no cache", false},
+		{"cached", true},
+	}
+	for _, tc := range tcs {
+		b.Run(tc.name, func(b *testing.B) {
+			certificates.PemCacheEnabled = tc.enabled
```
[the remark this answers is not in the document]
xopoww: I understand this point, but it seems that introducing struct type and so on only for this test is not ideal (it brings more complexity and potentially reduces test coverage). I've already [tried](https://github.com/xopoww/ydb-go-sdk/blob/ff166e0dba9ccbf74dd67328a8fd167e04f77eb1/internal/cache/cache.go) introducing similar solution and we decided not to do it.

How about instead using `(*testing.T) Cleanup()` to reset all global settings to how they were before the test?

internal/certificates/certificates.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+// ParseCertificatesFromFile reads and parses pem-encoded certificate(s) from file.
+func ParseCertificatesFromFile(file string) ([]*x509.Certificate, error) {
+	if !FileCacheEnabled {
+		certs, err := parseCertificatesFromFile(file)
+		return certs, xerrors.WithStackTrace(err)
+	}
+
+	value, exists := fileCache.Load(file)
```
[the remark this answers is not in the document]
xopoww: Yes, the problem is possible. We have another cache `string(decoded pem block) -> *x509.Certificate` which seems like what you're talking about. Caching by filename saves some resources compared to only PEM cache (as indicated by benchmark).

Possible solutions (as I see them) are:

 - remove caching by filename at all and leave only PEM cache
 - delete filename cache records on come condition (e.g. on certificate expiration time `cert.NotAfter`) + maybe give users an option of manually forcing uncached certificate loading

Caching by **file** contents / hash is not good because the most expensive operation is `os.ReadFile`. I don't have a benchmark testing this kind of caching, but I did compare `os.ReadFile` with `pem.Decode`, and the former is much more expensive.

certificates_test.go
```diff
@@ -0,0 +1,186 @@
...
+	"testing"
+
+	"github.com/ydb-platform/ydb-go-sdk/v3/internal/certificates"
+	"github.com/ydb-platform/ydb-go-sdk/v3/internal/conn"
+)
+
+func BenchmarkWithCertificateCache(b *testing.B) {
+	b.ReportAllocs()
+
+	bytes, err := os.ReadFile(os.Getenv("YDB_SSL_ROOT_CERTIFICATES_FILE"))
```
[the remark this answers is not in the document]
xopoww: I could add a `testdata/` directory and put a root CAs file from the current version of ydb container there. This will also improve the reproducibility of tests/benchmarks, which is also good.

certificates_test.go
xopoww: This test is in root package because we need to benchmark and test `connection.With` method because resource usage in this method is the main motivation for the issue. Testing for example just `certificates.ParseCertificatesFromFile` will give much less information.

internal/certificates/certificates.go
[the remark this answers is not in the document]
xopoww: We could also use `os.Lstat()` and `fileInfo.ModTime()` to determine if the file was modified after caching. Here is a comparison between `os.ReadFile` and `os.Lstat`:

```
BenchmarkFileStuff/os.ReadFile-8         	  341007	      3507 ns/op	    1080 B/op	       5 allocs/op
BenchmarkFileStuff/os.Lstat___-8         	 1000000	      1081 ns/op	     272 B/op	       2 allocs/op
```

internal/certificates/certificates.go
xopoww: The main problem is determining the suitable removal policy. For PEM cache the obvious solution is to remove expired (i.e. `time.Now() > cert.NotAfter`) certificates from cache. However, in applications with a lot of certificates with long expiration periods memory usage can still be big. For filename cache it is less obvious because the cached value is a slice of certs.

Implementing something smart like LRU/LFU seems to be an overkill in this situation, but I might be wrong here. We could also do something like FIFO with max cache records limit.

Need more info on real-world applications from which came the motivation for the issue.

certificates_test.go
```diff
@@ -0,0 +1,186 @@
...
+				if err != nil {
+					b.Fatal(err)
+				}
+			}
+		})
+	}
+}
+
+func TestWithCacheConcurrent(t *testing.T) {
+	const nRoutines = 10
```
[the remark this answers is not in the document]
xopoww: Done

internal/certificates/certificates.go
xopoww: Done

asmyasnikov:
You need to rebase on master (or merge) and than re-push your branch

certificates_test.go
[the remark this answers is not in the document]
rekby: I think global code and global state is bad.

func globalReset() may be possible workaround, but better use struct with methods.

@asmyasnikov, what you think?

internal/certificates/certificates.go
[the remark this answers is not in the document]
rekby: In your example you can read file 285000 times per second.
How often do you read certificates in your case?

internal/certificates/certificates.go
rekby: I think it will not be a big problem in real life, but the cache need any way to cleaning for prevent memory leak in rare cases.

certificates_test.go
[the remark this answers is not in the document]
rekby: ydb container generate new tmp certificates on every create.

If you will put fixed certificate to fixed path - it must have a long time before expire (100 years for example).

certificates_test.go
```diff
@@ -0,0 +1,186 @@
...
+func BenchmarkWithCertificateCache(b *testing.B) {
+	b.ReportAllocs()
+
+	bytes, err := os.ReadFile(os.Getenv("YDB_SSL_ROOT_CERTIFICATES_FILE"))
+	if err != nil {
+		b.Fatal(err)
+	}
+
+	// these vars make applyOptions() do unnecessary things
+	os.Unsetenv("YDB_SSL_ROOT_CERTIFICATES_FILE")
```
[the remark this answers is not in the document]
rekby: Yes, restore env in cleanup is good way for the test.

certificates_test.go
rekby: Now find the file is difficult and when I see to certificates.go - I see about it has no tests.

If you want solve problem of cycle import - you can declare package with _test suffix (```package certificates_test```) in certificates_test.go, placed near to certificates.go

internal/certificates/certificates.go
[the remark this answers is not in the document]
xopoww: Currently filename cache saves about 4.6 us and 2.3 KB per child connection (call to `db.With`):

```
BenchmarkWithFileCache/no_cache-8                  37642             32658 ns/op           44066 B/op        247 allocs/op
BenchmarkWithFileCache/cert_cache-8                43017             27734 ns/op           41731 B/op        210 allocs/op
BenchmarkWithFileCache/both_caches-8               52866             23121 ns/op           39379 B/op        191 allocs/op
```

So I think that expensiveness of `ReadFile`/`Lstat` should be compared to these values, not just by itself.

Unfortunately, I don't have enough info on real-life cases related to the issue. Maybe, the original issuer might provide some details.

certificates_test.go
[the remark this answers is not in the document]
xopoww: Now the benchmarks don't actually connect to YDB, so for them the expiration date of certificate doesn't really matter. The tests, however, do connect to the YDB, but it also means that they require additional work in a form of starting YDB container (like other e2e tests in this project).

I could rewrite the tests like benchmarks, but I'm afraid it might make concurrency test less informative (we won't check for concurrent cert access during concurrent child connections to database).

with.go
```diff
@@ -31,6 +31,19 @@ func (c *connection) With(ctx context.Context, opts ...Option) (Connection, erro
...
 		)...,
 	)
+	if err != nil {
+		return nil, 0, xerrors.WithStackTrace(err)
+	}
+	return child, id, nil
+}
+
+func (c *connection) With(ctx context.Context, opts ...Option) (Connection, error) {
+	child, id, err := c.with(ctx, opts...)
```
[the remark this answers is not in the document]
xopoww: Used [here](https://github.com/xopoww/ydb-go-sdk/blob/1f64baf6d78cb66ebbb52b5cc13cd38093be9ccf/with.go#L54). The separation between `conn.with` and `conn.With` looks a little ugly, but it is much better for the testing.

certificates_test.go
xopoww: Renamed test file to `with_test.go`

Issue #364, opened 2022-08-30
dev: Structured logger


rekby:
Ticket https://st.yandex-team.ru/KIKIMR-15742

Add interface for structural logger and move internal logger to it.

Next purpose - replace zap-logger, zerolog logger implementations to implement the interface and remove a lot of duplicate code for every logger.

PR #451, opened 2022-11-14
Structural logger draft

xopoww:

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: #364

## Other information

Satellite PR: https://github.com/ydb-platform/ydb-go-sdk-zap/pull/8 (zap implementation of structural logger)


rekby:
Hello, thanks for you work. I saw the work, it is beautifully code. It is good idea for comment Message as last call - for allow internal optimizations in adapters.

I saw some problems:
1. Now adapter must implement very many methods (summary about 50). It is a lot of code.
2. Most method duplicated between Record and Array
3. We have no way for extend the logger in the feature (for example if we will need new field type)
4. We have no way for improve logging from sdk side.
5. We need always prepare log values and call methods for chain even if logger is disabled for target level.

In my ideal world logger interface must be:
1. Zero cost if not logged.
2. Very cheap while used
3. Simple at implementation

I suggest few changes:
1. Simplify logger interface as much as possible, ideal case is single method. We don't need pass logger through function calls and can prepare all options before call logger. 
```go
type Logger interface {
    Enabled(level LogLevel) bool
    Log(opts Options, message string, fields ...Field)
}

type Options struct {
    Names []string
    Level LogLevel
}
```
It will allow fast stop prepare log messages, data, etc if logger disabled at current level. If logger enabled - we can send all values with one call, without chain calls.

It has less overhead: no call chains, no allocations for accumulate fields in implementations. Strict limits for live internal object time - it can be free before Log return.

2. Use common struct Field for all loggers with a methods for every interested type + universal fallback method - for allow add safe add new types in feature.
```go

type FieldType int 

const (
    UnknownType FieldType = iota
    IntType
    StringType
    ...
)

type Field struct {
    key string
    fieldType FieldType
    
    vint int
    vstring string
    ... 
    vinterface interface{}
}

func (f *Field) Key() string { return f.key }
func (f *Field) Type() FieldType
func (f *Field) Int() int { return f.vint }
func (f *Field) String() string { return f.vstring }
func (f *Field) Interface() interface{} { return f.v }
func (f *Field) SomeType() SomeType { return f.vinterface.(SomeType)}

// Universal methods - for simple implementations and feature compatibles
func (f *Field) UniversalString() string {
   switch f.fieldType {
   case IntType:
      return strconv.Itoa(f.vint)
    ...
   }
}

func f(f *Field) UniversalWriteStringBytes(w io.Writer) {
   switch f.fieldType {
    case IntType:
       buf := bytesPool.Get().(*[]byte) // <- get zero length, bytes with prepared capacity for no allocations
       *buf = strconv.AppendInt(*buf, f.vint, 10)
       _, _ = w.Write(*buf)
       *buf = *buf[:0]
       bytesPool.Put(buf)
    ...
   }
}
```

And we can add new field types and new optimizations, change internal store layout without break external adapters.

The changes allow implement very simple and very effective adapters for external loggers.

For example simplest logger:
```go
type PrintLogger struct {}

func(PrintLogger) Enabled(level structured.Level) bool {
  return True
}

func(PrintLogger) Log(opt structured.Options, message string, fields ...structured.Field){
    fmt.Print(strings.Join(opt.Names),": ", message)
    if len(fields) == 0 {
      return
    }

    last := field[len(fields)-1]
    fields = fields[:len(fields)-1]
    for _, field := range fields {
       fmt.Print(field.Key(), ":", field.UniversalString(), ", ")
    }
    fmt.Println(field.Key(), ":", field.UniversalString())
}
```

Issue #439, opened 2022-11-07
RFC: consumed units from metadata over context


asmyasnikov:
Main idea: add to context callback which will call on getting metadata headers
This solution can use for aggregate consumed units per call or for few calls at the same time

PR #453, opened 2022-11-15
Add handler for getting incoming metadata + refactoring

asmyasnikov:

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: #439 

## What is the new behavior?

-
-
-


version.go
```diff
@@ -1,6 +1,8 @@
 package ydb
 
-import "github.com/ydb-platform/ydb-go-sdk/v3/internal/meta"
+import (
+	"github.com/ydb-platform/ydb-go-sdk/v3/meta"
+)
 
 // Version reports current version of sdk
 const Version = meta.Version
```
rekby, line 8: Now version available in two places:
1. There
2. meta/version.go

I think it should be removed from public available in meta/version.go

sugar/consumed_units.go
```diff
@@ -0,0 +1,39 @@
...
+		for _, v := range values {
+			consumedUnits, err := strconv.ParseUint(v, 10, 64)
+			if err == nil {
+				atomic.AddUint64(&c.consumedUnits, consumedUnits)
+			}
+		}
+	})
+}
+
+func (c *consumedUnitsCounter) Sum() uint64 {
```
rekby, line 33: What about Sum -> Get?

sugar/consumed_units.go
```diff
@@ -0,0 +1,39 @@
...
+	return atomic.LoadUint64(&c.consumedUnits)
+}
+
+func (c *consumedUnitsCounter) Clear() {
```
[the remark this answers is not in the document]
asmyasnikov: 1. For reuse consumedUnitsCounter on other call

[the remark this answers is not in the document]
asmyasnikov: 2. Ack

sugar/consumed_units.go
asmyasnikov: Ack

version.go
asmyasnikov: You are right. But splitting version helps to exclude imports each other. Root package cant be imported from sub-packages because it provokes mutual imports and code cant be compiled.
Root version stayed for backward compatibility
meta/version.go - for newest usage. May be I didnt right

internal/table/session.go
```diff
@@ -218,10 +223,8 @@ func (s *session) KeepAlive(ctx context.Context) (err error) {
 		onDone(err)
 	}()
 
-	t := s.trailer()
-	defer t.processHints()
 	resp, err := s.tableService.KeepAlive(
-		balancer.WithEndpoint(ctx, s),
+		balancer.WithEndpoint(incoming.WithMetadataCallback(ctx, s.checkCloseHint), s),
```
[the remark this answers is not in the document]
asmyasnikov: Good think bor another PR

version.go
rekby: What about - hide generated version to internal (as in main)?

I doesn't like many deprecated code with renaming, backward compatible and so on.
It make more difficult to understand public interface.

Rename/move possible, if it has strong reason only.

asmyasnikov: Ack

PR #454, opened 2022-11-15
Moved `types.NewTimeToLiveSettings` to internal/table/session.go

asmyasnikov:
Defined `types.NewTimeToLiveSettings` constructor with options

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-


table/options/options.go
```diff
@@ -112,8 +114,21 @@ func (settings timeToLiveSettings) ApplyCreateTableOption(d *CreateTableDesc, a
 	d.TtlSettings = (*TimeToLiveSettings)(&settings).ToYDB()
 }
 
-func WithTimeToLiveSettings(settings TimeToLiveSettings) CreateTableOption {
-	return timeToLiveSettings(settings)
+// WithTTL defines TTL settings in CreateTable request
+func WithTTL(columnName string, ttl time.Duration, opts ...timeToLiveSettingsOption) CreateTableOption {
```
rekby, line 118: I suggest add function WithTimeToLiveSettings only - for uniformity.
Other functions no double variants, it receive settings object only.

table/options/options.go
```diff
@@ -608,33 +623,76 @@ type (
 	}
 )
 
+// WithAddColumn adds column in AlterTable request
+//
+// Deprecated: use AddColumn instead
```
rekby, line 628: Why you want remove prefix With from some of functions?

other functions exists with the prefix, for example: With Alter Partition Settings Object, With Alter Key Bloom Filter, ... and other options in other packages.

I think prefix removing should be:
1. Separate PR with rename/deprecate only
2. Rename all public options
3. Think about complete remove prefix, or change it, for example: Alter Table Add Column, Alter Table Set TTL, ProfileReplicationPolicy, CreateTableBloomFilter, etc - for uniform naming and prevent names clash.

table/options/options.go
```diff
@@ -608,33 +623,76 @@ type (
...
+func AddColumn(name string, typ types.Type) AlterTableOption {
+	return column{
+		name: name,
+		typ:  typ,
+	}
+}
+
+// WithAlterAttribute changes attribute in AlterTable request
+//
+// Deprecated: use AlterAttribute instead
```
rekby, line 646: Even if deprecate old names - replace it by call right method.
It will prevent sync functional in the future and prompt right usage within code.

PR #474, opened 2022-12-02
zap-like structural logger

xopoww:

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: #364


logs/field.go
```diff
@@ -0,0 +1,211 @@
...
+
+func String(key string, value string) Field {
+	return Field{
+		Type: StringType,
+		Key:  key,
+		vstr: value,
+	}
+}
+
+func (f Field) String() string {
```
xopoww, line 44: Maybe getters in Field should be called `GetX` instead of `X`? E.g.
```go
func (f Field) GetString() string {...}
```
This will make it easier to specify the purpose of such methods. Also, it will remove unwanted associations: now type `Field` implements `fmt.Stringer`, but semantically `func (Field) String() string` has a very different meaning.

logs/field.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+}
+
+// Error is the same as NamedError("error", value)
+func Error(value error) Field {
+	return NamedError("error", value)
+}
+
+func (f Field) Error() error {
+	return f.vany.(error)
```
xopoww, line 121: This line will panic `if f.vany == nil`. To guard against it I specified in the comment that getter `X` **must not** be called if `f.Type != XType`. Need to verify if this is ok behavior for SDK.

internal/logs/traces/discovery.go
```diff
@@ -0,0 +1,62 @@
...
+	t.OnDiscover = func(info trace.DiscoveryDiscoverStartInfo) func(trace.DiscoveryDiscoverDoneInfo) {
+		ll.Info("discover start",
+			logs.String("address", info.Address),
+			logs.String("database", info.Database),
+		)
+		start := time.Now()
+		return func(info trace.DiscoveryDiscoverDoneInfo) {
+			if info.Error == nil {
+				ll.Debug("discover done",
+					logs.Duration("latency", time.Since(start)),
```
xopoww, line 30: Several fields (latency, version) are very common between all traces. Maybe we should add something like
```go
func latency(start) logs.Field { return logs.Duration("latency", time.Since(start)) }

func version() logs.Field { return logs.String("version", meta.Version) }
```
in this package to DRY? The drawback of this is implicit behavior, which is not a big problem here IMO.

logs/field.go
asmyasnikov: Get* is a java and c# like naming. In go it is unusual naming as I know
zapcore.Field have public fields. This is unsafed way, because fields are have read write access.
Why need use String() as fmt.Stringer interface? In which cases?

logs/field.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+func nilField(key string) Field {
+	return Field{
+		Type: NilType,
+		Key:  key,
+	}
+}
+
+func Endpoints(key string, value []trace.EndpointInfo) Field {
```
rekby, line 148: I think we should no add internal structures as field types.

If somebody with adapter X will "optimize" output endpoints, then we will add new field for our trace.EndpointInfo: we will no see new field in the logs from adapter X.

All convert from internal stuctures to plain types will in our internal logger.

logs/field.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+}
+
+func (f Field) Endpoints() []trace.EndpointInfo {
+	return f.vany.([]trace.EndpointInfo)
+}
+
+// Fallback returns default string representation of Field value.
+// It should be used by adapters that don't support f.Type directly.
+func (f Field) Fallback() string {
```
rekby, line 162: If we add prefix/suffix for get values - the method can be rename to String() - for implement Stringer interface.

logs/field.go
```diff
@@ -0,0 +1,211 @@
...
+
+	// StringsType corresponds to []string
+	StringsType
+
+	ErrorType
+	// AnyType indicates that the Field is untyped. Adapters should use
+	// reflection-based approached to marshal this field.
+	AnyType
+	// NilType indicates that the Field value is nil.
+	NilType
```
rekby, line 207: I think no need "NilType" - it can be store only one value and it no need for logs.

And no "EndpointsType" - we need to convert all internal structs to primitive fields internally.

internal/logs/traces/discovery.go
rekby: about "version":

I think it can be common func-helper with "auto-add" version field, use pool for create fields slices, may be select log level, similar to [logDebugInfo](https://github.com/ydb-platform/ydb-go-sdk-zap/blob/35ffa3a654f6a8bffb6b31901fa8f60be2f0eece/loglevel.go#L12-L12) helper in current zap adapter (used for topic).

@asmyasnikov what do you think about the helper?

logs/logger.go
```diff
@@ -0,0 +1,10 @@
+package logs
+
+type Logger interface {
+	Log(opts Options, msg string, fields ...Field)
```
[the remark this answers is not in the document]
xopoww: How should this Enabled(level) be used in SDK? I'm not sure it is needed because logger design is already almost zero-cost for disabled levels. Checking if level is enabled in traces will make the traces much more complicated, while most (good) adapters are already checking it and guaranteeing almost zero-cost for disabled logs.

It is not a problem to add this method to the Logger iface, I'm just not sure why.

logs/field.go
xopoww: The problem here is that `[]trace.EndpointInfo` is essentially `[]fmt.Stringer`, but it cannot be converted to that type without reallocations/reflection. So adding internal (exported, to be exact) types as FieldType is just a way to make logger as zero-cost as possible.

Zap [solves](https://pkg.go.dev/go.uber.org/zap#Stringers) this by using generics, but it is not an option if we want to support users with older Go versions.

logs/field.go
xopoww: NilType is added to handle cases when, for example, user passes nil error to `logs.Error()`. There are two ways of handling this problem: a) ignore this field entirely or b) marshal it as `"<nil>"` or something like that. By adding a special type for these cases we make both options possible (adapter decides how to handle nil fields).

logs/field.go
xopoww: Yes, @rekby is right about what I mean. I like the idea of renaming `f.String() => f.StringValue()` because `GetString` really does not look good in Go.

logs/logger.go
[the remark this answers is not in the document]
rekby: For example topic API can receive messages in a batch, it can contains many small messages per one network package (up to 50-60 messages).
For 10Mbit/second stream it will up to 35 000 messages, for 100Mbit/second - up to 350 000 messages.

If customer will receive messages one by one - near zero cost may be visible. It can cause some records for every message in max level (start wait message, receive message + may be some internal states), but it will be filtered in most cases be log level filter.

I think tracker com check Enabled if need (on very hot path or if prepare have some cost). The tracer may check or may not check the level.

Logger adapter:
1. Can implement enabled check (return true - for stub).
2. Can be called even if enabled return false or not called at all.

It is way for optimization only.

logs/field.go
rekby: We can create type endpoints
```
type Endpoints []trace.EndpointInfo
```

local to tracer package and implement stringer for Endpoints (not for every endpoints). Convert from []trace.EndpointInfo to Endpoints is cheap and no need allocations.

logs/field.go
rekby: We can story AnyType as type field and return nil by ```Any()``` method, these options still available and code simple.

If we start check nil it cause next questions: will we need check if any is string, int, ... ?

logs/field.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+	return f.vany.(map[string][]string)
+}
+
+// Fallback returns default string representation of Field value.
+// It should be used by adapters that don't support f.Type directly.
+// Non-nil error indicates some fatal marshaling error and is most
+// likely a sign of invalid Field initialization.
+func (f Field) Fallback() (repr string, err error) {
+	func() {
```
rekby, line 115: It is no need func outer to defer.

logs/field.go
xopoww: It is a solution, but it makes slice of endpoints to always be one field with no internal structure, whereas now adapters can make it a nested array of fields if they support it. If we marshal endpoints to text, we need to choose how to marshal it and take away this choice from user/adapter/etc. For example, we can stringify `[]trace.EndpointInfo` as JSON array, but what if someone wants to use some other markup language for log messages?

We could also do something like
```go
type Marshaler interface {
    json.Marshaler
    xml.Marshaler
    ...
}

func MarshalerField(key string, value Marshaler) Field {...}
```
But it just replaces one **fixed** format with several **fixed** formats.

internal/logs/traces/discovery.go
xopoww: Seems like this functionality could be added to [logger](https://github.com/xopoww/ydb-go-sdk/blob/364-logs/internal/logs/traces/logger.go) from `internal/logs/traces`.

logs/logger.go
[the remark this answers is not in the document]
xopoww: Added `Enabled()` method.

logs/field.go
xopoww: Replaced internal types with `fmt.Stringer`.

logs/field.go
xopoww: Replaced `NilType` with `AnyType` and `f.vany = nil.`

logs/field.go
xopoww: Fixed

logs/field.go
xopoww: Need final decision on this naming issue. `f.Foo() -> f.FooValue()` & `f.Fallback() -> f.String()`?

internal/logs/traces/types.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/ydb-platform/ydb-go-sdk/v3/trace"
+)
+
+type endpoints []trace.EndpointInfo
+
+func (es endpoints) String() string {
+	bldr := strings.Builder{}
+	bldr.WriteByte('[')
+	for i, e := range es {
+		bldr.WriteString(e.String())
```
xopoww, line 17: Need to think on how we marshal internal structures. For example, do we need to json-escape this string (given that any JSON encoder in logger adapter will also escape it, resulting in double escaping like `","` -> `"\,"` -> `"\\\,"`)?

Just using `json.Marshal(es)` does not work, unfortunately. json package does not use Stringer's `String` method for marshaling.

logs/field.go
rekby: I think f.Fallback() -> f.String() is good and I like f.FooValue() for get values.  

fmt.Stringer is very userful interface and often it used implicity, it is very easy to write fmt.Println(f) and it should work good by my opinion.

logs/field.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+// String is a value getter for fields with StringType type
+func (f Field) String() string {
+	return f.vstr
+}
+
+// Int is a value getter for fields with IntType type
+func (f Field) Int() int {
+	return int(f.vint)
```
rekby, line 50: what about check type for Int and other type-specified value getters?
And panic if use it for bad type.

For example if we want use interface{} for all values in future - we will do f.vany.(int) and it will panic if bool stored.
And it will consistent behavior for all values. Now it has different behaviour between mix int/bool and int/strings values.

logs/field.go
```diff
@@ -0,0 +1,260 @@
...
+func (f Field) Error() error {
+	if f.vany == nil {
+		return nil
+	}
+	return f.vany.(error)
+}
+
+// Any is a value getter for fields with AnyType type
+func (f Field) Any() interface{} {
+	return f.vany
```
rekby, line 85: What about return values with known type?
For example return bool for BoolType?

logs/field.go
```diff
@@ -0,0 +1,260 @@
...
+// It should be used by adapters that don't support f.Type directly.
+// Non-nil error indicates some fatal marshaling error and is most
+// likely a sign of invalid Field initialization.
+func (f Field) Fallback() (repr string, err error) {
+	defer func() {
+		if p := recover(); p != nil {
+			if e, ok := p.(error); ok {
+				err = e
+			} else {
+				err = fmt.Errorf("%v", p)
```
rekby, line 106: what if use: "ydb: panic with ..." and wrap the error with xerrors.Wrap + xerrors.WithStacktrace?

logs/field.go
```diff
@@ -0,0 +1,260 @@
...
+	// StringsType corresponds to []string
+	StringsType
+
+	ErrorType
+	// AnyType indicates that the Field is untyped. Adapters should use
+	// reflection-based approached to marshal this field.
+	AnyType
+
+	// StringerType corresponds to fmt.Stringer
+	StringerType
```
rekby, line 259: What if add MaxType as last item? - fot test purposes

internal/logs/traces/topic.go
```diff
@@ -0,0 +1,450 @@
+// //nolint:lll
+package traces
+
+// import (
```
rekby, line 4: Why the code commented?

internal/logs/traces/types.go
rekby: we can implement https://pkg.go.dev/encoding#TextMarshaler for the endpoint.info - it will used by json.Marshal.

log/adapter.go
```diff
@@ -0,0 +1,103 @@
...
+		f = logger.Debugf
+	case logs.INFO:
+		f = logger.Infof
+	case logs.WARN:
+		f = logger.Warnf
+	case logs.ERROR:
+		f = logger.Errorf
+	case logs.FATAL:
+		f = logger.Fatalf
+	default:
```
rekby, line 34: Need explicit return for quiet level only and some error for unknown level

log/adapter.go
```diff
@@ -0,0 +1,103 @@
...
+				format = "<error:%q>"
+				value = err
+			} else {
+				format = "%s"
+				value = fb
+			}
+		}
+		fmt.Fprintf(fmtBldr, `%q:%s`, field.Key(), format)
+		if i != len(fields)-1 {
+			fmtBldr.WriteByte(',')
```
rekby, line 89: what about space after comma?

log/adapter.go
```diff
@@ -0,0 +1,103 @@
...
+		case logs.AnyType:
+			format = "%v"
+			value = field.Any()
+		case logs.StringerType:
+			format = "%s"
+			value = field.Stringer().String()
+		default:
+			if fb, err := field.Fallback(); err != nil {
+				format = "<error:%q>"
+				value = err
```
rekby, line 81: err -> err.Error() ?

log/adapter.go
```diff
@@ -0,0 +1,103 @@
...
+		var (
+			format string
+			value  interface{}
+		)
+		//nolint:goconst
+		switch field.Type() {
+		case logs.InvalidType:
+			continue
+		case logs.IntType:
+			format = "%d"
```
rekby, line 52: What about use %v for all types?

internal/logs/traces/fields.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"time"
+
+	"github.com/ydb-platform/ydb-go-sdk/v3/internal/meta"
+	"github.com/ydb-platform/ydb-go-sdk/v3/logs"
+)
+
+// latency creates Field "latency": time.Since(start)
+func latency(start time.Time) logs.Field {
```
rekby, line 11: great!

internal/logs/traces/table.go
```diff
@@ -0,0 +1,777 @@
...
+					logs.Duration("latency", time.Since(start)),
+					logs.Bool("idempotent", idempotent),
+				)
+			} else {
+				f := ll.Warn
+				if !xerrors.IsYdb(info.Error) {
+					f = ll.Debug
+				}
+				m := retry.Check(info.Error)
+				f(`do attempt failed`,
```
rekby, line 46: What about use fixed place for error? Always first or always last - for better find it while read logs?

logs/field.go
```diff
@@ -0,0 +1,260 @@
...
+	case StringType:
+		repr = f.vstr
+	case BoolType:
+		repr = strconv.FormatBool(f.Bool())
+	case DurationType:
+		repr = f.Duration().String()
+	case StringsType:
+		repr = fmt.Sprintf("%v", f.Strings())
+	case ErrorType:
+		repr = f.Error().Error()
```
rekby, line 124: It will panic if err == nil.

logs/field.go
```diff
@@ -0,0 +1,260 @@
...
+		return nil
+	}
+	return f.vany.(fmt.Stringer)
+}
+
+// Fallback returns default string representation of Field value.
+// It should be used by adapters that don't support f.Type directly.
+// Non-nil error indicates some fatal marshaling error and is most
+// likely a sign of invalid Field initialization.
+func (f Field) Fallback() (repr string, err error) {
```
rekby, line 100: Need table test for fallback representation for all value types.

logs/field.go
```diff
@@ -0,0 +1,260 @@
...
+		key:   key,
+		vany:  value,
+	}
+}
+
+// NamedError constructs Field with ErrorType. If value is nil,
+// resulting Field will be of AnyType instead of ErrorType.
+func NamedError(key string, value error) Field {
+	if value == nil {
+		return Any(key, nil)
```
rekby, line 198: What about use ErrorType for nil error.

It allow skip nil errors in adapter and not show error fields for messages without error (as zap)

internal/logs/traces/table.go
```diff
@@ -0,0 +1,777 @@
...
+		)
+		start := time.Now()
+		return func(info trace.TableDoIntermediateInfo) func(trace.TableDoDoneInfo) {
+			if info.Error == nil {
+				ll.Trace(`do attempt done`,
+					logs.Duration("latency", time.Since(start)),
+					logs.Bool("idempotent", idempotent),
+				)
+			} else {
+				f := ll.Warn
```
rekby, line 41: What about direct use log method with level as argument instead of dinamic select proxy-function?

log/adapter.go
```diff
@@ -0,0 +1,103 @@
...
+			value = field.Strings()
+		case logs.ErrorType:
+			format = "%v"
+			value = field.Error()
+		case logs.AnyType:
+			format = "%v"
+			value = field.Any()
+		case logs.StringerType:
+			format = "%s"
+			value = field.Stringer().String()
```
rekby, line 77: What about send field.Stringer() value as is - without render to string right now?
formatter can render it later (if need) or skip render it log record will skip later (for example logger.Trace method will simple return)

asmyasnikov:
Maybe need to use golang.org/x/exp/slog interface ?

Issue #494, opened 2022-12-22
dev: unknown errors as undefined for retry reason


rekby:
Current behaviour: All unknown codes threaded as Finished, which deny retry of them.
Suggested behaviour: Mark known errors as Finished/NotFinished/Undefined and return Undefined state for all unknown codes.

[code](https://github.com/ydb-platform/ydb-go-sdk/blob/488-retry-policy/internal/xerrors/transport.go#L107-L107)

PR #496, opened 2022-12-22
Refactoring of internal retry types

asmyasnikov:

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #494

## What is the new behavior?

-
-
-


internal/xerrors/operation.go
```diff
@@ -101,19 +100,19 @@ func IsOperationError(err error, codes ...Ydb.StatusIds_StatusCode) bool {
...
 		Ydb.StatusIds_BAD_SESSION,
 		Ydb.StatusIds_SESSION_BUSY:
-		return operation.NotFinished
+		return TypeRetryable
 	case Ydb.StatusIds_UNDETERMINED:
-		return operation.Undefined
+		return TypeConditionallyRetryable
 	default:
-		return operation.Finished
+		return TypeNonRetryable
```
rekby, line 115: What about explicit add known unretriable operations as separate case and use default as fallback only.

internal/xerrors/type.go
```diff
@@ -0,0 +1,28 @@
+package xerrors
+
+import "fmt"
+
+// Type reports which error type
+type Type uint8
+
+const (
+	TypeNoError = Type(iota)
```
rekby, line 9: What about iota + 1 or add TypeUndefined as zero item?

for use explicit values only?

internal/xsql/conn.go
```diff
@@ -139,7 +138,7 @@ func (c *conn) PrepareContext(ctx context.Context, query string) (_ driver.Stmt,
 
 func (c *conn) execContext(ctx context.Context, query string, args []driver.NamedValue) (_ driver.Result, err error) {
 	m := queryModeFromContext(ctx, c.defaultQueryMode)
-	onDone := trace.DatabaseSQLOnConnExec(c.trace, &ctx, query, m.String(), retry.IsIdempotent(ctx))
+	onDone := trace.DatabaseSQLOnConnExec(c.trace, &ctx, query, m.String(), xerrors.IsIdempotent(ctx))
```
rekby, line 141: What about move IsIdempotent, WithIdempotent from xerrors to xcontext?
```
xcontext.IsIdempotent(ctx)
```

asmyasnikov: ack

internal/xerrors/operation.go
asmyasnikov: ack

Issue #507, opened 2023-01-09
Add support for Xorm in SDK


datbeohbbh:
See: [#7](https://github.com/ydb-platform/xorm/issues/7#issue-1525047333)

PR #508, opened 2023-01-09
Add support for Xorm in SDK

datbeohbbh:

## Pull request type
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: [#507](https://github.com/ydb-platform/ydb-go-sdk/issues/507#issue-1525096703)

### DIFF
--- a/internal/xsql/conn.go
+++ b/internal/xsql/conn.go
@@ -216,12 +216,12 @@ func (c *conn) QueryContext(ctx context.Context, query string, args []driver.Nam
 	if c.isClosed() {
 		return nil, errClosedConn
 	}
-	if c.currentTx != nil {
-		return c.currentTx.QueryContext(ctx, query, args)
-	}
-	if _, ok := ctx.Value(XormMetadataQuery).(string); ok {
+	if checkXormQueryFromContext(ctx) {
 		xorm := newXorm(ctx, c, query, args)
 		return xorm.queryMetadata()
+	}
+	if c.currentTx != nil {
+		return c.currentTx.QueryContext(ctx, query, args)
 	}
 	return c.queryContext(ctx, query, args)
 }
--- a/internal/xsql/conn.go
+++ b/internal/xsql/conn.go
@@ -219,6 +219,10 @@ func (c *conn) QueryContext(ctx context.Context, query string, args []driver.Nam
 	if c.currentTx != nil {
 		return c.currentTx.QueryContext(ctx, query, args)
 	}
+	if _, ok := ctx.Value(XormMetadataQuery).(string); ok {
+		xorm := newXorm(ctx, c, query, args)
+		return xorm.queryMetadata()
+	}
 	return c.queryContext(ctx, query, args)
 }
 
--- a/internal/xsql/rows.go
+++ b/internal/xsql/rows.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"database/sql/driver"
 	"io"
+	"strings"
 	"sync"
 
 	"github.com/ydb-platform/ydb-go-sdk/v3/internal/xerrors"
@@ -15,9 +16,10 @@ import (
 )
 
 var (
-	_ driver.Rows              = &rows{}
-	_ driver.RowsNextResultSet = &rows{}
-	_ driver.Rows              = &single{}
+	_ driver.Rows                           = &rows{}
+	_ driver.RowsNextResultSet              = &rows{}
+	_ driver.RowsColumnTypeDatabaseTypeName = &rows{}
+	_ driver.Rows                           = &single{}
 
 	_ types.Scanner = &valuer{}
 )
@@ -42,6 +44,33 @@ func (r *rows) Columns() []string {
 		i++
 	})
 	return cs
+}
+
+// Add `ColumnTypeDatabaseTypeName` to support `Xorm`
+// NOTE: Need to optimize somehow? This might take O(|r.Columns()|^2)
+// each time (*Rows).ColumnTypes be called
+// https://cs.opensource.google/go/go/+/refs/tags/go1.19.4:src/database/sql/sql.go;l=3101
+// https://cs.opensource.google/go/go/+/refs/tags/go1.19.4:src/database/sql/sql.go;drc=f721fa3be9bb52524f97b409606f9423437535e8;l=3141
+func (r *rows) ColumnTypeDatabaseTypeName(index int) string {
+	r.nextSet.Do(func() {
+		r.result.NextResultSet(context.Background())
+	})
+
+	f := func(s string) string {
+		if strings.HasPrefix(s, "Optional") {
+			s = strings.TrimPrefix(s, "Optional<")
+			s = strings.TrimSuffix(s, ">")
+		}
+		return strings.ToUpper(s)
+	}
+
+	var i int
+	yqlTypes := make([]string, r.result.CurrentResultSet().ColumnCount())
+	r.result.CurrentResultSet().Columns(func(m options.Column) {
+		yqlTypes[i] = f(m.Type.Yql())
+		i++
+	})
+	return yqlTypes[index]
 }
 
 func (r *rows) NextResultSet() error {
--- a/internal/xsql/connector.go
+++ b/internal/xsql/connector.go
@@ -8,6 +8,7 @@ import (
 	metaHeaders "github.com/ydb-platform/ydb-go-sdk/v3/internal/meta"
 	"github.com/ydb-platform/ydb-go-sdk/v3/internal/xerrors"
 	"github.com/ydb-platform/ydb-go-sdk/v3/meta"
+	"github.com/ydb-platform/ydb-go-sdk/v3/scheme"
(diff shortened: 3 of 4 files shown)


internal/xsql/rows.go
```diff
@@ -44,6 +45,25 @@ func (r *rows) Columns() []string {
 	return cs
 }
 
+// Add `ColumnTypeDatabaseTypeName` to support `Xorm`
+// NOTE: Need to optimize somehow? This might take O(|r.Columns()|^2)
```
asmyasnikov, line 49: may be first call of r.result.CurrentResultSet().Columns() need to store column types to internal rows cache?
We get O(|r.Columns()|+1)

no need to add some cache now, it is not a problem. But wrong implementation of cache can have bugs, difficult case - behaviour on iterate to next result set

internal/xsql/conn.go
```diff
@@ -377,56 +377,80 @@ func (c *conn) ResetSession(_ context.Context) error {
...
-		metadataQuery          = "metadata-query"
-		versionQueryMode       = "version"
-		isTableExistQueryMode  = "is-table-exist"
-		isColumnExistQueryMode = "is-column-exist"
-		getColumnsQueryMode    = "get-columns"
-		getTablesQueryMode     = "get-tables"
-		getIndexesQueryMode    = "get-indexes"
-	)
-
+func (c *conn) Version(ctx context.Context) (_ driver.Rows, err error) {
```
asmyasnikov, line 380: need to add conn_test.go with simple typecheck
```go
var _ interface {
  Version(context.Context) (result)
} = (*conn)(nil)
```

asmyasnikov, line 380: Return type - string (1) or strings (2) (major, minor, patch) and error
(1) is simple and better

asmyasnikov: because at now YDB cannot return self version. 
Now need to return "default" for example

asmyasnikov: You need to know next info: YDB is a distributed RDBMs and version - is a versions of all nodes YDB. 
For example some nodes have 22-4 version, some - 22-2, some - 22-5. And its O'k

internal/xsql/conn.go
```diff
@@ -377,56 +377,80 @@ func (c *conn) ResetSession(_ context.Context) error {
...
-	if !ok {
-		return nil, fmt.Errorf("context is not in metadata query mode")
+	return &single{
+		values: []sql.NamedArg{
+			sql.Named("Version", "YDB Server "+"v"+version),
+		},
+	}, nil
+}
+
+func (c *conn) IsTableExists(ctx context.Context, query string, args []driver.NamedValue) (_ driver.Rows, err error) {
```
asmyasnikov, line 395: see previous comment about test for fix signature for next refactorings

internal/xsql/conn.go
```diff
@@ -377,56 +377,80 @@ func (c *conn) ResetSession(_ context.Context) error {
...
+		},
+	}
+	err = mr.isTableExist(ctx, query, args)
+	if err != nil {
+		return nil, err
 	}
+	return mr, nil
+}
 
+func (c *conn) IsColumnExists(ctx context.Context, query string, args []driver.NamedValue) (_ driver.Rows, err error) {
```
asmyasnikov, line 408: see previous comment about test for fix signature for next refactorings

internal/xsql/conn.go
```diff
@@ -377,82 +381,310 @@ func (c *conn) ResetSession(_ context.Context) error {
...
-		},
+func (c *conn) GetColumnType(ctx context.Context, tableName, columnName string) (dataType string, err error) {
+	var (
+		cn      = c.connector.Connection()
+		session = c.session
+	)
+
+	tableExist, err := helpers.IsTableExists(ctx, cn.Scheme(), tableName)
+	if err != nil {
+		return
```
asmyasnikov, line 469: Returned nil bc err shadowed in 467

asmyasnikov: And bottom the same mistake

internal/xsql/conn.go
```diff
@@ -365,28 +365,18 @@ func (c *conn) BeginTx(ctx context.Context, txOptions driver.TxOptions) (_ drive
...
 func (c *conn) Version(ctx context.Context) (_ string, err error) {
-	const version = "default"
+	const version string = "default"
 	return version, nil
 }
 
 func (c *conn) IsTableExists(ctx context.Context, tableName string) (tableExists bool, err error) {
-	var cn = c.connector.Connection()
+	cn := c.connector.Connection()
 	tableExists, err = helpers.IsTableExists(ctx, cn.Scheme(), tableName)
```
asmyasnikov, line 375: return helpers.IsTableExists(ctx, cn.Scheme(), tableName) ?

internal/xsql/conn.go
```diff
@@ -365,28 +365,18 @@ func (c *conn) BeginTx(ctx context.Context, txOptions driver.TxOptions) (_ drive
...
 	if err != nil {
-		return
+		return false, err
```
asmyasnikov, line 377: or return false, xerrors.WithStackTrace(err)

internal/xsql/conn.go
```diff
@@ -513,44 +503,44 @@ func (c *conn) GetPrimaryKeys(ctx context.Context, tableName string) (pkCols []s
...
 		if pkCol == columnName {
 			ok = true
 			break
 		}
 	}
-	return
+	return ok, nil
 }
 
 func (c *conn) GetTables(ctx context.Context, root string) (tables []string, err error) {
```
asmyasnikov, line 546: absPath - maybe root or some child dir inside root

internal/xsql/conn_test.go
```diff
@@ -0,0 +1,45 @@
...
+	_ interface {
+		Version(context.Context) (string, error)
+	} = (*conn)(nil)
+
+	_ interface {
+		IsTableExists(context.Context, string) (bool, error)
+	} = (*conn)(nil)
+
+	_ interface {
+		IsColumnExists(context.Context, string, string) (bool, error)
```
asmyasnikov, line 15: add var names for explicit understanding of API

internal/xsql/conn.go
```diff
@@ -360,3 +364,311 @@ func (c *conn) BeginTx(ctx context.Context, txOptions driver.TxOptions) (_ drive
...
+		if _, ignore := ignoreDirs[dir]; ignore {
+			return false
+		}
+		return true
+	}
+
+	queue := make([]string, 0)
+	queue = append(queue, root)
+
+	for st := 0; st < len(queue); st++ {
```
asmyasnikov, line 567: I think no need to iterate over subdirs. This is feature is not allowed in any RDBMs, only YDB as I know. If some app with xorm need to restrict access with some directory - this code must be define table path prefix explicitely.

asmyasnikov: If we will want to get all tables including embedding tables in subdirs - we add code like this

asmyasnikov: At now it is overengenering

internal/xsql/conn_test.go
asmyasnikov: and result names

internal/xsql/rows.go
```diff
@@ -48,6 +49,23 @@ func (r *rows) Columns() []string {
...
+// each time (*Rows).ColumnTypes be called
+// https://cs.opensource.google/go/go/+/refs/tags/go1.19.4:src/database/sql/sql.go;l=3101
+func (r *rows) ColumnTypeDatabaseTypeName(index int) string {
+	r.nextSet.Do(func() {
+		r.result.NextResultSet(context.Background())
+	})
+
+	var i int
+	yqlTypes := make([]string, r.result.CurrentResultSet().ColumnCount())
```
asmyasnikov, line 61: it must be a some cache field of rows. for second call

asmyasnikov: Or at now it look like overengenering, bc making slice and call .Yql() is not free

asmyasnikov (commented):
See details in PR

internal/xsql/conn.go
```diff
@@ -362,3 +366,354 @@ func (c *conn) BeginTx(ctx context.Context, txOptions driver.TxOptions) (_ drive
...
 	return c.currentTx, nil
 }
+
+func (c *conn) Version(ctx context.Context) (_ string, err error) {
+	const version string = "default"
+	return version, nil
+}
+
+func (c *conn) IsTableExists(ctx context.Context, tableName string) (tableExists bool, err error) {
+	cn := c.connector.Connection()
```
asmyasnikov, line 376: Please replace all `cn` vars to `cc`. For single style of code

asmyasnikov (commented):
PR have small issues from me

internal/xsql/conn.go
datbeohbbh: @asmyasnikov I fixed, please check again

asmyasnikov (approved):
LGFM

PR #530, opened 2023-01-23
Drop `internal/xnet` + replace permanent connection into newest connection each discovery

asmyasnikov:

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-


internal/balancer/balancer.go
```diff
@@ -71,7 +65,12 @@ func (b *Balancer) clusterDiscovery(ctx context.Context) (err error) {
 		)
 	}()
 
-	endpoints, err = b.discovery.Discover(ctx)
+	client := b.discoveryClient()
+	if err != nil {
```
TheHurd, line 69: what error are we checking in this place? (where did you get it from?)

internal/discovery/discovery.go
```diff
@@ -20,24 +20,31 @@ import (
...
-	config  config.Config
-	service Ydb_Discovery_V1.DiscoveryServiceClient
-	cc      conn.Conn
+	config      config.Config
+	grpcOptions []grpc.DialOption
+}
+
+func (c *Client) client(ctx context.Context) (Ydb_Discovery_V1.DiscoveryServiceClient, *grpc.ClientConn, error) {
+	cc, err := grpc.DialContext(ctx,
+		"dns:///"+c.config.Endpoint(),
```
TheHurd, line 41: it is better to do such things once, when parsing the configuration, and not every time you connect

internal/discovery/discovery.go
```diff
@@ -65,7 +72,15 @@ func (c *Client) Discover(ctx context.Context) (endpoints []endpoint.Endpoint, e
 		return nil, xerrors.WithStackTrace(err)
 	}
 
-	response, err = c.service.ListEndpoints(ctx, &request)
+	client, cc, err := c.client(ctx)
+	if err != nil {
+		return nil, xerrors.WithStackTrace(err)
+	}
+	defer func() {
+		_ = cc.Close()
```
TheHurd, line 80: here and in other similar places maybe it's better to make a wrapper for close() to apply a simpler defer form?

internal/balancer/balancer.go
asmyasnikov: ack

internal/discovery/discovery.go
asmyasnikov: simple form have linter issue (errcheck)

hurd54: that's why I suggested making a wrapper for close()
something like:

```go
func (c someType) silentClose {
  _ = c.Close()
}

...
// somewhere in the code
defer cc.silentClose()
```

PR #579, opened 2023-02-15
Issue364 structlog

tgnike:

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: https://github.com/ydb-platform/ydb-go-sdk/issues/364

## What is the new behavior?

- re-review of (https://github.com/ydb-platform/ydb-go-sdk/pull/474)


logs/field.go
```diff
@@ -0,0 +1,260 @@
...
+func (f Field) Type() FieldType {
+	return f.ftype
+}
+
+func (f Field) Key() string {
+	return f.key
+}
+
+// String is a value getter for fields with StringType type
+func (f Field) String() string {
```
rekby, line 44: What about rename String, Int, ... to StringValue, IntValue, etc?

It allow rename Fallback to String and implement Stringer interface for comfortable usage Field in formatters

logs/field.go
```diff
@@ -0,0 +1,260 @@
...
+	return f.vstr
+}
+
+// Int is a value getter for fields with IntType type
+func (f Field) Int() int {
+	return int(f.vint)
```
rekby, line 50: What about check f.Type and panic if call Int() for bad time (for example String).

That allow to see bad values used and will consistent with other types combination, for example Strings and Stringer - it will panic while cast interface.

log/adapter.go
```diff
@@ -0,0 +1,103 @@
...
+	case logs.INFO:
+		f = logger.Infof
+	case logs.WARN:
+		f = logger.Warnf
+	case logs.ERROR:
+		f = logger.Errorf
+	case logs.FATAL:
+		f = logger.Fatalf
+	default:
+		return
```
rekby, line 35: Need explicit return for quiet level and some error for unknown level as default

log/adapter.go
```diff
@@ -0,0 +1,103 @@
...
+		case logs.AnyType:
+			format = "%v"
+			value = field.Any()
+		case logs.StringerType:
+			format = "%s"
+			value = field.Stringer().String()
+		default:
+			if fb, err := field.Fallback(); err != nil {
+				format = "<error:%q>"
+				value = err
```
rekby, line 81: err -> err.Error() ?

log/adapter.go
```diff
@@ -0,0 +1,103 @@
...
+	for i, field := range fields {
+		var (
+			format string
+			value  interface{}
+		)
+		//nolint:goconst
+		switch field.Type() {
+		case logs.InvalidType:
+			continue
+		case logs.IntType:
```
rekby, line 51: What about use %v for all types?

internal/logs/traces/table.go
```diff
@@ -0,0 +1,777 @@
...
+			} else {
+				f := ll.Warn
+				if !xerrors.IsYdb(info.Error) {
+					f = ll.Debug
+				}
+				m := retry.Check(info.Error)
+				f(`do attempt failed`,
+					logs.Duration("latency", time.Since(start)),
+					logs.Bool("idempotent", idempotent),
+					logs.Error(info.Error),
```
rekby, line 49: What about use fixed place for error? Always first or always last - for better find it while read logs?

logs/field.go
```diff
@@ -0,0 +1,260 @@
...
+	case StringType:
+		repr = f.vstr
+	case BoolType:
+		repr = strconv.FormatBool(f.Bool())
+	case DurationType:
+		repr = f.Duration().String()
+	case StringsType:
+		repr = fmt.Sprintf("%v", f.Strings())
+	case ErrorType:
+		repr = f.Error().Error()
```
rekby, line 124: It will panic if err == nil.

internal/logs/traces/table.go
```diff
@@ -0,0 +1,777 @@
...
+		)
+		start := time.Now()
+		return func(info trace.TableDoIntermediateInfo) func(trace.TableDoDoneInfo) {
+			if info.Error == nil {
+				ll.Trace(`do attempt done`,
+					logs.Duration("latency", time.Since(start)),
+					logs.Bool("idempotent", idempotent),
+				)
+			} else {
+				f := ll.Warn
```
rekby, line 41: What about direct use log method with level as argument instead of dinamic select proxy-function?

log/adapter.go
```diff
@@ -0,0 +1,103 @@
...
+			value = field.Strings()
+		case logs.ErrorType:
+			format = "%v"
+			value = field.Error()
+		case logs.AnyType:
+			format = "%v"
+			value = field.Any()
+		case logs.StringerType:
+			format = "%s"
+			value = field.Stringer().String()
```
rekby, line 77: What about send field.Stringer() value as is - without render to string right now?
formatter can render it later (if need) or skip render it log record will skip later (for example logger.Trace method will simple return)

log/adapter.go
rekby: And need test for all levels

logs/field.go
```diff
@@ -41,72 +41,97 @@ func (f Field) Key() string {
...
 	return f.vany.(error)
 }
 
 // Any is a value getter for fields with AnyType type
-func (f Field) Any() interface{} {
-	return f.vany
+func (f Field) AnyValue() interface{} {
+	f.checkType(AnyType)
+
+	switch f.vany.(type) {
```
rekby, line 94: you can skip the switch and simple return f.vany with same result.

logs/field.go
```diff
@@ -257,4 +278,12 @@ const (
 
 	// StringerType corresponds to fmt.Stringer
 	StringerType
+
+	EndType
 )
+
+func (ft FieldType) String() string {
+
+	return []string{"invalid", "int", "int64", "string", "bool", "time.Duration", "[]string", "error", "any", "stringer", "endtype"}[ft]
```
rekby, line 287: 1. It is not efficient - create []string slice for every function call
2. It is bad readable and bad checkable (if we reorder fields in definition)

What about use switch and case for every variant?

logs/field_test.go
```diff
@@ -0,0 +1,79 @@
...
+			tests = append(tests, test{name: name, f: Any(name, 9223372036854775807), want: "9223372036854775807"})
+			tests = append(tests, test{name: name, f: Any(name, "any string"), want: "any string"})
+			tests = append(tests, test{name: name, f: Any(name, true), want: "true"})
+			tests = append(tests, test{name: name, f: Any(name, []string{"Abc", "Def", "Ghi"}), want: "[Abc Def Ghi]"})
+			tests = append(tests, test{name: name, f: Any(name, errors.New("error")), want: "error"})
+			tests = append(tests, test{name: name, f: Any(name, nil), want: "<nil>"})
+		case StringerType:
+			tests = append(tests, test{name: name, f: Stringer(name, &stringerTest{}), want: "stringerTest"})
+		default:
+			tests = append(tests, test{name: "panic " + name, f: Field{ftype: ftype, key: "default"}, want: "", panic: true})
```
rekby, line 63: 1. Why not use common pattern for table test?

```go
tests := []struct{
		name  string
 		f     Field
 		want  string
 		panic bool
}{
  {
    name: "int",
    f: Int("int", 1),
    want: "1",
  }
}
```

name field can be skipped and use field name as test name


2. in that style: why not t.Fail right in default case?

tgnike: 1. This makes easier to track changes in types
2. Ok

log/adapter_test.go
```diff
@@ -0,0 +1,159 @@
...
+		},
+		{
+			name: "Unknown level",
+			a:    logger,
+			args: args{opts: logs.Options{Lvl: logs.Level(99)}, msg: "Unknown level", fields: fields},
+			want: `unknown level`,
+			fail: true,
+		},
+	}
+	for _, tt := range tests {
```
rekby, line 145: There is no check about all types was checked.

What about make ```types  := map[FieldType]bool{}``` and for every field mark ```types[field. ftype] = true```, then check about all types from 0 to EndType was present?

logs/field_test.go
rekby: need check about all types was checked same as above (map + for from 0 to EndType)

logs/field.go
```diff
@@ -0,0 +1,321 @@
...
+func (f Field) ErrorValue() error {
+	f.checkType(ErrorType)
+	if f.vany == nil {
+		return nil
+	}
+	return f.vany.(error)
+}
+
+// Any is a value getter for fields with AnyType type
+func (f Field) AnyValue() interface{} {
```
rekby, line 91: Need test for AnyValue()

tgnike: in this commit
https://github.com/ydb-platform/ydb-go-sdk/pull/579/commits/2ecccc30ae59b64f9116f0e4fea1827b77bc9447

logs/field_test.go
```diff
@@ -59,21 +62,71 @@ func TestField_String(t *testing.T) {
...
+		{name: "any_int64", f: Any("any", 9223372036854775807), want: 9223372036854775807},
+		{name: "any_string", f: Any("any", "any string"), want: "any string"},
+		{name: "any_bool", f: Any("any", true), want: true},
+		{name: "any_[]string", f: Any("any", []string{"Abc", "Def", "Ghi"}), want: []string{"Abc", "Def", "Ghi"}},
+		{name: "any_error", f: Any("any", errors.New("error")), want: errors.New("error")},
+		{name: "struct", f: Any("any", struct{ str string }{str: "test"}), want: struct{ str string }{str: "test"}},
+		{name: "any_nil", f: Any("any", nil), want: nil},
+	}
+
+	for _, tt := range tests {
```
rekby, line 125: In the test need check about all types verified too.

It is important because unexpected type will panic.

logs/field_test.go
```diff
@@ -22,15 +21,19 @@ func TestField_String(t *testing.T) {
...
 		panic bool
+		fail  bool
 	}
 
+	types := make(map[FieldType]bool)
+
 	tests := []test{}
 
-	for i := 0; i < int(EndType); i++ {
+	for i := 0; i <= int(EndType); i++ {
```
rekby, line 31: EndType must be not included - it isn't real type, it is marker only and the EndType value will be never use in program.

logs/field_test.go
tgnike: There is default section, test will fail with unexpected type

Issue #811, opened 2023-09-07
feat: Batch read requests for topic reader


rekby:
# Feature Request

**Describe the Feature Request**
Now topic reader send readrequest after every message/batch, returned from sdk. Need batch read requests for reduce requests count and allow server to send larger batches.

**Describe Preferred Solution**
Don't send read request right after return batch, but wait while returned data size reach 30% of buffer size.

**Related Code**
[send read requests](https://github.com/ydb-platform/ydb-go-sdk/blob/5a33c7b816652280ca9a3cb66b44c373772c99f4/internal/topic/topicreaderinternal/stream_reader_impl.go#L544-L544)
[free buffer after return message](https://github.com/ydb-platform/ydb-go-sdk/blob/5a33c7b816652280ca9a3cb66b44c373772c99f4/internal/topic/topicreaderinternal/stream_reader_impl.go#L200-L200)

**Additional Context**

**If the feature request is approved, would you be willing to submit a PR?**
Yes / No _(Help can be provided if you need assistance submitting a PR)_

PR #821, opened 2023-09-11
Feature/batch read requests for topic reader

gingersamurai:

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: #811


internal/topic/topicreaderinternal/stream_reader_impl.go
```diff
@@ -541,29 +543,40 @@ func (r *topicStreamReaderImpl) dataRequestLoop(ctx context.Context) {
...
 			}
 		}
 	}
 }
 
+var errCannotCalcFreeSpacePercentage = errors.New("cannot calculate free space percentage")
+
+func (r *topicStreamReaderImpl) getFreeSpacePercentage() (percentage int, err error) {
+	sizeBytes := float64(r.cfg.BufferSizeProtoBytes)
+	usedSpaceBytes := float64(r.getRestBufferBytes())
```
rekby, line 571: getRestBufferBytes return free bytes in buffer, not used.


what about rename the function?

and the value is bad for request data, because in show real buffer usage (decrease when receive data from server and increase when return data from sdk).
It will race conditions and request a lot of data over buffer size.

need another counter for reflect bytes, which sdk can request from server. It is not same as real usage.

the getRestBufferBytes usable for monitoring purposes only.

PR #853, opened 2023-10-07
return concrete type instead of interface for topic client

rekby:

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Driver was return interface type for topic client.
It will problem in future, when we need to extend the client.

# New behaviour
Driver return pointer to struct instead of interface.


asmyasnikov:
https://github.com/ydb-platform/ydb-go-sdk/discussions/907

PR #1006, opened 2024-01-18
Fix topic writer infinite reconnections

art22m:

Hello! 

I've created topic writer with `StartTimeout` 30 seconds and Default or Retry `RetryPolicy`.
Then with ip6tables I've blocked the YDB port to get transport error.
The retries are expected to stop after 30 seconds, but this does not happened. 

The reason why this is happened is `connectionTimeout*resetAttemptEmpiricalCoefficient` overflow. 
If `connectionTimeout` is `time.Duration(math.MaxInt64)` (that is always true, since we do not have options to set `connectionTimeout`), then its multiplication to empirical coefficient (=10) gives negative value (see playground https://goplay.space/#qlIeS6o3PCz). 
As a result, function `CheckResetReconnectionCounters` always returns true. Since `CheckResetReconnectionCounters` gives true, `startOfRetries` in `connectionLoop` loop method always sets to `w.clock.Now()`. Thus, in `CheckRetryMode` there are no chance to get `retriesDuration > settings.StartTimeout` to stop reconnections. As a result, we always get infinite reconnections in `Retry` and `Default` modes.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Topic writer with `StartTimeout` set to X and Default or Retry `RetryPolicy`.
Topic writer gets any error in Retry mode or retryable error in Default mode. 
After X time topic writer **continue** reconnections. 

Issue Number: N/A

## What is the new behavior?
Topic writer with `StartTimeout` set to X and Default or Retry `RetryPolicy`.
Topic writer gets any error in Retry mode or retryable error in Default mode. 
After X time topic writer **stop** reconnections. 

## Other information

Condition `startOfRetries.IsZero()` is needed to set `startOfRetries` for the first time.
Since with infinite connection duration `CheckResetReconnectionCounters` will return false.

### DIFF
--- a/internal/topic/retriable_error.go
+++ b/internal/topic/retriable_error.go
@@ -13,7 +13,8 @@ import (
 )
 
 const (
-	DefaultStartTimeout = time.Minute
+	DefaultStartTimeout          = time.Minute
+	connectionEstablishedTimeout = time.Minute
 )
 
 type RetrySettings struct {
@@ -46,7 +47,7 @@ var (
 func CheckResetReconnectionCounters(lastTry, now time.Time, connectionTimeout time.Duration) bool {
 	const resetAttemptEmpiricalCoefficient = 10
 	if connectionTimeout == value.InfiniteDuration {
-		return false
+		return now.Sub(lastTry) > connectionEstablishedTimeout
 	}
 	return now.Sub(lastTry) > connectionTimeout*resetAttemptEmpiricalCoefficient
 }
--- a/internal/topic/retriable_error.go
+++ b/internal/topic/retriable_error.go
@@ -93,9 +93,10 @@ func CheckRetryMode(err error, settings RetrySettings, retriesDuration time.Dura
 		return nil, false
 	}
 
-	if mode.BackoffType() == backoff.TypeFast {
+	switch mode.BackoffType() {
+	case backoff.TypeFast:
 		return backoff.Fast, true
+	default:
+		return backoff.Slow, true
 	}
-
-	return backoff.Slow, true
 }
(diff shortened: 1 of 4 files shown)


rekby:
@art22m thanks for your PR, I have been see it and will check soon in details.

Fix [broken test](https://github.com/ydb-platform/ydb-go-sdk/actions/runs/7570744399/job/20717850468?pr=1006#step:4:78) please.

rekby:
Hello.

Thanks for your pr and a lot of work for research and describe the bug.

Function CheckResetReconnectionCounters used for detect when we have established stream and reset counters - for good logging.
The PR solve bug with stop reconnection, but break logging/metric ux.

What you mean about set some reasonable constant for check "connection established" state for case with connection timeout is infinite?

For example - 1 minute. It will be mean:
if last connetion attemp was earler, than 1 minute - then connection was fine established and worked. And reset attemp counter to zero.

art22m:
Hi, thanks for feedback.
> if last connetion attemp was earler, than 1 minute - then connection was fine established and worked. And reset attempt counter to zero.

May be I did not get your point, but should we check last connection was earlier then constant, not later?

I guess you want to compare constant with lastConnectionAttempt variable. 
If the connection was fine stablished, and it worked for some time (`reconnectReason = writer.WaitClose(ctx)`), then lastConnectionAttempt will be larger then our constant and we should zero the attempts.
Contrariwise, if we do not established our connection, this variable always will be small due to this: `prevAttemptTime = now`

rekby:
When reconnect timeout is infinite, then we should check: if duration since last attempt was more then constant.

I want detect a situation:
1. Successfully establish a stream session
2. Successfully receive init message
3. Failed on fist (or one of firsts) message

And count the failure as new attempt for logs and for retry policy.

art22m:
I've made one minute constant, but I guess right constant should be find empirically.

rekby:
@art22m Thanks for the fix:)

PR #1126, opened 2024-03-13
Enable funlen linter

brojeg:

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting, renaming)
- [ ] Build related changes

Issue Number: [934](https://github.com/ydb-platform/ydb-go-sdk/issues/934)

## What is the new behavior?

-
-
-

### DIFF
--- a/internal/bind/numeric_args.go
+++ b/internal/bind/numeric_args.go
@@ -5,6 +5,7 @@ import (
 	"strconv"
 	"unicode/utf8"
 
+	"github.com/ydb-platform/ydb-go-sdk/v3/internal/params"
 	"github.com/ydb-platform/ydb-go-sdk/v3/internal/xerrors"
 	"github.com/ydb-platform/ydb-go-sdk/v3/internal/xstring"
 	"github.com/ydb-platform/ydb-go-sdk/v3/table"
@@ -16,9 +17,7 @@ func (m NumericArgs) blockID() blockID {
 	return blockYQL
 }
 
-func (m NumericArgs) RewriteQuery(sql string, args ...interface{}) (
-	yql string, newArgs []interface{}, err error,
-) {
+func (m NumericArgs) RewriteQuery(sql string, args ...interface{}) (yql string, newArgs []interface{}, err error) {
 	l := &sqlLexer{
 		src:        sql,
 		stateFn:    numericArgsStateFn,
@@ -29,14 +28,18 @@ func (m NumericArgs) RewriteQuery(sql string, args ...interface{}) (
 		l.stateFn = l.stateFn(l)
 	}
 
-	var (
-		buffer = xstring.Buffer()
-		param  table.ParameterOption
-	)
+	buffer := xstring.Buffer()
 	defer buffer.Free()
 
 	if len(args) > 0 {
-		newArgs = make([]interface{}, len(args))
+		parameters, err := parsePositionalParameters(args)
+		if err != nil {
+			return "", nil, err
+		}
+		newArgs = make([]interface{}, len(parameters))
+		for i, param := range parameters {
+			newArgs[i] = param
+		}
 	}
 
 	for _, p := range l.parts {
@@ -49,38 +52,23 @@ func (m NumericArgs) RewriteQuery(sql string, args ...interface{}) (
 			}
 			if int(p) > len(args) {
 				return "", nil, xerrors.WithStackTrace(
-					fmt.Errorf("%w: $p%d, len(args) = %d", ErrInconsistentArgs, p, len(args)),
+					fmt.Errorf("%w: $%d, len(args) = %d", ErrInconsistentArgs, p, len(args)),
 				)
 			}
-			paramName := "$p" + strconv.Itoa(int(p-1)) //nolint:goconst
-			if newArgs[p-1] == nil {
-				param, err = toYdbParam(paramName, args[p-1])
-				if err != nil {
-					return "", nil, xerrors.WithStackTrace(err)
-				}
-				newArgs[p-1] = param
-				buffer.WriteString(param.Name())
-			} else {
-				buffer.WriteString(newArgs[p-1].(table.ParameterOption).Name())
+			paramIndex := int(p - 1)
+			if newArgs[paramIndex] == nil {
+				return "", nil, xerrors.WithStackTrace(ErrInconsistentArgs)
 			}
+			buffer.WriteString(newArgs[paramIndex].(table.ParameterOption).Name())
 		}
 	}
 
-	for i, p := range newArgs {
-		if p == nil {
-			return "", nil, xerrors.WithStackTrace(
-				fmt.Errorf("%w: $p%d, len(args) = %d", ErrInconsistentArgs, i+1, len(args)),
-			)
-		}
-	}
-
+	yql = buffer.String()
 	if len(newArgs) > 0 {
-		const prefix = "-- origin query with numeric args replacement\n"
-
-		return prefix + buffer.String(), newArgs, nil
+		yql = "-- origin query with numeric args replacement\n" + yql
 	}
 
-	return buffer.String(), newArgs, nil
+	return yql, newArgs, nil
 }
 
 func numericArgsStateFn(l *sqlLexer) stateFn {
@@ -128,6 +116,20 @@ func numericArgsStateFn(l *sqlLexer) stateFn {
 			return nil
 		}
 	}
+}
+
+func parsePositionalParameters(args []interface{}) ([]*params.Parameter, error) {
+	newArgs := make([]*params.Parameter, len(args))
+	for i, arg := range args {
+		paramName := fmt.Sprintf("$p%d", i)
+		param, err := toYdbParam(paramName, arg)
+		if err != nil {
+			return nil, err
+		}
+		newArgs[i] = param
+	}
+
+	return newArgs, nil
 }
 
 func numericArgState(l *sqlLexer) stateFn {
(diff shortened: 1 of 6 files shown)


brojeg:
@rekby Please review.

internal/bind/numeric_args.go
```diff
@@ -49,38 +52,23 @@ func (m NumericArgs) RewriteQuery(sql string, args ...interface{}) (
...
-				param, err = toYdbParam(paramName, args[p-1])
-				if err != nil {
-					return "", nil, xerrors.WithStackTrace(err)
-				}
-				newArgs[p-1] = param
-				buffer.WriteString(param.Name())
-			} else {
-				buffer.WriteString(newArgs[p-1].(table.ParameterOption).Name())
+			paramIndex := int(p - 1)
+			if newArgs[paramIndex] == nil {
```
rekby, line 59: It will be never true in new code: newArgs fulled by `parsePositionalParameters` and if it has some error - the function will return earlier.

internal/bind/params.go
```diff
@@ -174,57 +174,44 @@ func toYdbParam(name string, value interface{}) (*params.Parameter, error) {
...
+			if pErr != nil {
+				return nil, xerrors.WithStackTrace(pErr)
 			}
-			parameters = append(parameters, param)
+			newParams = append(newParams, param)
+		}
+		if err != nil {
+			return nil, err
+		}
+		if parameters == nil {
```
rekby, line 211: no need the condition, append work well with `nil` as first value:

```go
parameters = append(parameters, newParam)
```

internal/bind/params.go
```diff
@@ -233,3 +220,33 @@ func Params(args ...interface{}) (parameters []*params.Parameter, _ error) {
 
 	return parameters, nil
 }
+
+func handleNamedValue(arg driver.NamedValue, i, argsLen int) ([]*params.Parameter, error) {
```
rekby, line 224: 1. The function receive and return exactly one parameter, fix return value from `[]*params.Parameter` to `*params.Parameter`
2. Better function name: paramHandleNamedValue, with main function prefix. It is better understand about it helper function only.
3. It needs better name for argument `i`, for example `paramNumber`, because `i` hasn't counter semantic as argument.

internal/bind/params.go
```diff
@@ -174,57 +174,44 @@ func toYdbParam(name string, value interface{}) (*params.Parameter, error) {
 	return params.Named(name, v), nil
 }
 
-func Params(args ...interface{}) (parameters []*params.Parameter, _ error) {
-	parameters = make([]*params.Parameter, 0, len(args))
+func Params(args ...interface{}) ([]*params.Parameter, error) {
+	parameters := make([]*params.Parameter, 0, len(args))
 	for i, arg := range args {
+		var newParams []*params.Parameter
```
rekby, line 180: on every step the loop add only one parameter, change `newParams []*params.Parameter` to `newParameter *params.Parameter `.

internal/bind/params.go
```diff
@@ -174,57 +174,44 @@ func toYdbParam(name string, value interface{}) (*params.Parameter, error) {
...
-			}
+			newParams, err = handleNamedValue(x, i, len(args))
 		case sql.NamedArg:
 			if x.Name == "" {
 				return nil, xerrors.WithStackTrace(errUnnamedParam)
 			}
-			param, err := toYdbParam(x.Name, x.Value)
-			if err != nil {
-				return nil, xerrors.WithStackTrace(err)
+			param, pErr := toYdbParam(x.Name, x.Value)
```
rekby, line 189: here you can direct assign

```
newParam, err = toYdbParam(x.Name, x.Value)
```

Why do you need `pErr`? use existed `err` variable

internal/bind/params.go
```diff
@@ -174,57 +174,44 @@ func toYdbParam(name string, value interface{}) (*params.Parameter, error) {
...
+			if pErr != nil {
```
rekby, line 190: no need handle local errors at every case - you handle err after switch.

internal/bind/params.go
```diff
@@ -174,57 +174,44 @@ func toYdbParam(name string, value interface{}) (*params.Parameter, error) {
...
-				return nil, xerrors.WithStackTrace(err)
+			param, pErr := toYdbParam(fmt.Sprintf("$p%d", i), x)
+			if pErr != nil {
+				return nil, xerrors.WithStackTrace(pErr)
 			}
-			parameters = append(parameters, param)
+			newParams = append(newParams, param)
+		}
+		if err != nil {
+			return nil, err
```
rekby, line 209: use xerrors.WithStackTrace for wrap err

internal/conn/conn.go
```diff
@@ -392,13 +391,13 @@ func (c *conn) NewStream(
 ) (_ grpc.ClientStream, err error) {
```
rekby, line 391: it is better don't split the function:
the extracted function has many arguments and many return values. It isn't simple "createStream" it:
1. Create stream.
2. Create and return send mark
3. Create and return traceID

It is difficult to name one common purpose of new function because it make few operations.

internal/scripting/client.go
```diff
@@ -218,26 +218,53 @@ func (c *Client) StreamExecute(
 	return r, xerrors.WithStackTrace(err)
 }
 
+func receiveStreamResult(
```
rekby, line 221: the function doesn't receive result, it make function for receive result.

Better to rename to, for example to `newReceiveStreamResultFunc`

internal/scripting/client.go
```diff
@@ -218,26 +218,53 @@ func (c *Client) StreamExecute(
...
+	stream Ydb_Scripting_V1.ScriptingService_StreamExecuteYqlClient,
+	onIntermediate func(error) func(error),
```
rekby, line 223: `onIntermediate` better rename to `traceErr`

rekby (changes requested):
Thank, it looks much better, then previous pr.

Fix comments please and resolve merge conflicts.

internal/bind/params.go
brojeg: I'm not quite sure how to handle this comment. paramHandleNamedValueis returning *params.Parameter
How to properly handle case *params.Parameters: ?
If I'll return return (*x)[0] - the tests are failing with 
Error Trace:    params_test.go:526
                Error:          Not equal:
                                expected: []*params.Parameter{(*params.Parameter)(0xc000298b00), (*params.Parameter)(0xc000298b40), (*params.Parameter)(0xc000298b80)}
                                actual  : []*params.Parameter{(*params.Parameter)(0xc000298a40)}

rekby: you should change signature of `handleNamedValue` too - for return simple pointer: `*params.Parameter`, because `handleNamedValue` handle only one argument and can produce exactly one `*params.Parameter` or error. 

you can append the `newParam` to `parameters` after the switch.

Or I bad understand your question.

brojeg: I appreciate your feedback. However, I'm seeking some clarification on your suggestion to have the helper function return a *params.Parameter instead of a []*params.Parameter.

From my understanding, the original Params function is designed to handle multiple parameters ([]*params.Parameter) rather than a single one (*params.Parameter). This is evident in several parts of the function, including the handling of *params.Parameters and *params.Parameter types, where the function either assigns a slice of *params.Parameter or appends to it. The logic seems to be built around accumulating parameters in a slice, which is then sorted and returned.

Given this context, I'm concerned that changing the return type to *params.Parameter might not align with how the Params function is structured to accumulate and process multiple parameters. It might also affect the ability to directly append the result of the helper function to the parameters slice, as the function currently expects to work with slices throughout.

rekby: Original func `func Params(args ...interface{}) ([]*params.Parameter, error) {` - work with slice of params, but every single iteration of loop `for i, arg := range args {` work with only one parameter.

> I'm concerned that changing the return type to *params.Parameter might not align with how the Params function is structured to accumulate and process multiple parameters

new function - `func handleNamedValue(arg driver.NamedValue, i, argsLen int) ([]*params.Parameter, error) {` work as part of the loop, it take only one input parameter and return only one result parameter. It is same what do one loop iteration.

> It might also affect the ability to directly append the result of the helper function to the parameters slice, as the function currently expects to work with slices throughout.

it will not affect for append result to slice, because you can append single value:

```
parameters = append(parameters, newParams)
```

brojeg: I totally agree and do not have any objections on the  loop for `i, arg := range args {`  that is working with only one parameter. My question is about the operations with `parameters = make([]*params.Parameter, 0, len(args))`
On every iteration of the single `arg` in for loop we're facing `switch` statement - for the case  where out single parameter is type of` driver.NamedValue` - we have a couple options. 
if  `xx := x.Value.(type)` is type of `*params.Parameters` that is defined as (`type Parameters []*Parameter`) we're returning from this case statement  `parameters = *xx` (a slice of `[]*params.Parameter`) Not single `*params.Parameter` therefore the helper function will need to return` []*params.Parameter`.
The only way to return from helper function  single `*params.Parameter` that I can see is in the main Param function inside case `driver.NamedValue`: to keep `switch` statement with case `*params.Parameters`: and move inside the helper function only case `*params.Parameter`: and `default` behavior. But it looks a bit strange to me.

rekby: Ok, you convinced me to leave return type `[]*params.Parameter`

rekby:
Good work!

now it needs to resolve conflicts. 

And if you make startDiscovery be a method of Balancer struct it will be good, but it isn't necessary.

rekby:
@brojeg thanks for your work!

PR #1170, opened 2024-03-27
fix: scopelint linter

dlc-01:

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting, renaming)
- [x] Build related changes

Issue Number: #964 

## What is the new behavior?

-
-
-

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -241,7 +241,6 @@ linters:
     - maintidx
     - nonamedreturns
     - paralleltest
-    - scopelint
     - structcheck
     - testableexamples
     - testpackage
--- a/examples/ttl/series.go
+++ b/examples/ttl/series.go
@@ -268,6 +268,7 @@ func createTables(ctx context.Context, c table.Client, prefix string) (err error
 	}
 
 	for i := 0; i < expirationQueueCount; i++ {
+		i := i
 		err = c.Do(ctx,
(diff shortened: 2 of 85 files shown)


.golangci.yml
asmyasnikov, line 1: I think need add exclude rule for "*_test.go" files because major part of your PR about test files
Your fixes of test files doesn't really matter. Sorry

examples/ttl/series.go
```diff
@@ -268,6 +268,7 @@ func createTables(ctx context.Context, c table.Client, prefix string) (err error
 	}
 
 	for i := 0; i < expirationQueueCount; i++ {
+		i := i
```
asmyasnikov, line 271: You can store tableName as
```
tableName := path.Join(prefix, fmt.Sprintf("expiration_queue_%v", i)
```
And use tableName inside Do call

internal/cmd/gtrace/writer.go
```diff
@@ -402,6 +402,7 @@ func (w *Writer) composeHookCall(fn *Func, h1, h2 string) {
 				w.atEOL = true
 			}
 			for i, h := range []string{h1, h2} {
+				i, h := i, h
```
asmyasnikov, line 405:
```suggestion
```

PR #1268, opened 2024-06-10
* Added experimental range functions to the `query.Result` and `query.ResultSet` types, available starting with Go version 1.22. These features can be enabled by setting the environment variable `GOEXPERIMENT=rangefunc`.

asmyasnikov:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-


internal/query/result_range_test.go
```diff
@@ -0,0 +1,363 @@
...
+	stream.EXPECT().Recv().Return(nil, io.EOF)
+	r, _, err := newResult(ctx, stream, nil, nil)
+	require.NoError(t, err)
+	defer r.Close(ctx)
+	rsCount := 0
+	for rs, err := range r.Range(ctx) {
+		require.NoError(t, err)
+		rowsCount := 0
+		for _, err := range rs.Range(ctx) {
+			require.NoError(t, err)
```
rekby, line 355: Is it possible to compare row values with expected?

tests/integration/query_range_test.go
```diff
@@ -0,0 +1,176 @@
...
+		ydb.WithAccessTokenCredentials(os.Getenv("YDB_ACCESS_TOKEN_CREDENTIALS")),
+	)
+	require.NoError(t, err)
+	t.Run("Execute", func(t *testing.T) {
+		var (
+			p1 string
+			p2 uint64
+			p3 time.Duration
+		)
+		r, err := db.Query().Execute(ctx, `
```
rekby, line 40: Need test with more, then one line and more, then one result set - for check iterations.

internal/query/result_range_test.go
```diff
@@ -0,0 +1,363 @@
...
+
+	"github.com/stretchr/testify/require"
+	"github.com/ydb-platform/ydb-go-genproto/protos/Ydb"
+	"github.com/ydb-platform/ydb-go-genproto/protos/Ydb_Query"
+	"go.uber.org/mock/gomock"
+
+	"github.com/ydb-platform/ydb-go-sdk/v3/internal/xtest"
+)
+
+func TestResultRange(t *testing.T) {
```
rekby, line 18: need additional tests for behavior if user break loop.
I expect about result will be closed.

internal/query/result_set_range_test.go
```diff
@@ -0,0 +1,830 @@
...
+	"github.com/ydb-platform/ydb-go-genproto/protos/Ydb_Query"
+	"go.uber.org/mock/gomock"
+	grpcCodes "google.golang.org/grpc/codes"
+	grpcStatus "google.golang.org/grpc/status"
+
+	"github.com/ydb-platform/ydb-go-sdk/v3/internal/xerrors"
+	"github.com/ydb-platform/ydb-go-sdk/v3/internal/xtest"
+)
+
+func TestResultSetRange(t *testing.T) {
```
rekby, line 22: Need additional test for `break` behavior.
I expect to close the result set and I can iterate for next result set in the result.

internal/query/range.go
```diff
@@ -0,0 +1,16 @@
+//go:build !goexperiment.rangefunc
```
rekby, line 1: built tag doesn't contain go1.22 condition

internal/query/range_experiment.go
```diff
@@ -0,0 +1,55 @@
...
+	"context"
+	"io"
+	"iter"
+
+	"github.com/ydb-platform/ydb-go-sdk/v3/internal/xerrors"
+	"github.com/ydb-platform/ydb-go-sdk/v3/query"
+)
+
+type (
+	resultRange struct {
```
rekby, line 15: The struct same as for file without experimental. What if remove conditions for range.go and implement functions only within experimental func

internal/query/range_experiment.go
```diff
@@ -0,0 +1,55 @@
...
+		for {
+			rs, err := r.r.NextResultSet(ctx)
+			if err != nil {
+				if xerrors.Is(err, io.EOF) {
+					return
+				}
+			}
+			cont := yield(rs, err)
+			if !cont || err != nil {
+				return
```
rekby, line 34: Is it needs to close the result with defer?

internal/query/range_experiment.go
```diff
@@ -0,0 +1,55 @@
...
+		for {
+			rs, err := r.rs.NextRow(ctx)
+			if err != nil {
+				if xerrors.Is(err, io.EOF) {
+					return
+				}
+			}
+			cont := yield(rs, err)
+			if !cont || err != nil {
+				return
```
rekby, line 51: Is it needs to close the result with defer?

asmyasnikov: I think it is no need to do inside range func

internal/query/range_experiment.go
asmyasnikov: replaced with static funcs

PR #1307, opened 2024-06-24
use buffered channel to manage items in query pool

adwski:
## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

Current approach has couple shortcomings:
- items can be created above configured limit
- item will not be closed during shutdown if it was only in idle slice and not added to index map.
- pool has no warm-up, new items are created upon request 

## What is the new behavior?

I propose to use buffered channel to manage query pool items as oppose to index map and idle slice. This approach makes item creation process continuous and asynchronous, which allows seamlessly warm-up pool or re-establish all connections after connectivity failure.

Internally I introduced: 
- item queue (buffered channel) - main item queue
- item tokens (buffered channel) - item creation requests
- item spawner goroutine - runs asynchronously and fills queue with new items

getItem() and putItem() are working with item queue - receiving or sending back items. New thing about getItem() is that it ensures that retrieved item is ok to use (as far IsAlive() can tell). To achieve this it spins in endless loop until in can receive item from queue. This means in case of connectivity failure or pool starvation it will block causing back pressure to caller (who should use context with deadline if it wishes to not block as well but detect this situation).

### DIFF
--- a/internal/pool/defaults.go
+++ b/internal/pool/defaults.go
@@ -1,6 +1,22 @@
 package pool
 
-const DefaultLimit = 50
+import "time"
+
+const (
+	DefaultLimit = 50
+
+	// defaultCreateRetryDelay specifies amount of time that spawner will wait
+	// before retrying unsuccessful item create.
+	defaultCreateRetryDelay = 500 * time.Millisecond
+
+	// defaultSpawnGoroutinesNumber specifies amount of spawnItems goroutines.
+	// Having more than one spawner can potentially decrease warm-up time
+	// and connections re-establishment time after connectivity failure.
+	// Too high value will result in frequent connection establishment
+	// attempts (see defaultCreateRetryDelay) during connectivity
+	// issues which in some cases might not be desirable.
+	defaultSpawnGoroutinesNumber = 2
+)
 
 var defaultTrace = &Trace{
 	OnNew: func(info *NewStartInfo) func(info *NewDoneInfo) {
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -38,10 +38,26 @@ type (
 		createTimeout time.Duration
 		closeTimeout  time.Duration
 
-		mu    xsync.Mutex
-		idle  []PT
-		index map[PT]struct{}
-		done  chan struct{}
+		// queue is a buffered channel that holds ready-to-use items.
+		// Newly created items are sent to this channel by spawner goroutine.
+		// getItem reads from this channel to get items for usage.
+		// putItems sends item to this channel when it's no longer needed.
+		// Len of the buffered channel should be equal to configured pool size
+		// (MUST NOT be less).
+		// If item is in this queue, then it's considered idle (not in use).
+		queue chan PT
+
+		// itemTokens similarly to 'queue' is a buffered channel, and it holds 'tokens'.
+		// Presence of token in this channel indicates that there's requests to create item.
+		// Every token will eventually result in creation of new item (spawnItems makes sure of that).
+		//
+		// itemTokens must have same size as queue.
+		// Sum of every existing token plus sum of every existing item in any time MUST be equal
+		// to pool size. New token MUST be added by getItem/putItem if they discovered item in use to be
+		// no good and discarded it.
+		itemTokens chan struct{}
+
+		done chan struct{}
 
 		stats *safeStats
 	}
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+* Added warm-up and asynchronous item handling in internal/pool
```
asmyasnikov, line 1: no need to add a line for internal refactoring changes

asmyasnikov: Changelog for only user-facing changes

internal/pool/pool.go
```diff
@@ -172,16 +197,61 @@ func New[PT Item[T], T any](
...
+		case <-p.done:
+			break spawnLoop
+		case <-p.itemTokens:
+			// got token, must create item
+			for {
+				item, err := p.createItem(ctx)
+				if err != nil {
+					select {
+					case <-ctx.Done():
+						break spawnLoop
```
asmyasnikov, line 230: no need a label for loop - use return instead

internal/pool/pool.go
```diff
@@ -172,16 +197,61 @@ func New[PT Item[T], T any](
...
 
-	p.idle = make([]PT, 0, p.limit)
-	p.index = make(map[PT]struct{}, p.limit)
 	p.stats = &safeStats{
 		v:        stats.Stats{Limit: p.limit},
 		onChange: p.trace.OnChange,
 	}
 
+	for i := 0; i < defaultSpawnGoroutinesNumber; i++ {
+		go p.spawnItems(ctx)
```
asmyasnikov, line 206: context must be avalaible only on New call
After return managment to upper code - context maybe is done
For spawn goroutines - you can use xcontext.ValueOnly(ctx)

internal/pool/pool.go
```diff
@@ -172,16 +197,61 @@ func New[PT Item[T], T any](
...
+			// got token, must create item
+			for {
+				item, err := p.createItem(ctx)
+				if err != nil {
+					select {
+					case <-ctx.Done():
+						break spawnLoop
+					case <-p.done:
+						break spawnLoop
+					case <-time.After(defaultCreateRetryDelay):
```
asmyasnikov, line 233: time.After produce timer inside - this can be provokes the goroutines/rosurces leak
You can create ticker/timer before select - and call directly ticker.Stop() in defer

internal/pool/defaults.go
```diff
@@ -1,6 +1,22 @@
...
 
-const DefaultLimit = 50
+import "time"
+
+const (
+	DefaultLimit = 50
+
+	// defaultCreateRetryDelay specifies amount of time that spawner will wait
+	// before retrying unsuccessful item create.
+	defaultCreateRetryDelay = 500 * time.Millisecond
```
asmyasnikov, line 10: Can this setting helps for exactly once case?
I think this is over-engineering without the done cases

internal/pool/pool.go
```diff
@@ -358,25 +397,29 @@ func (p *Pool[PT, T]) putItem(ctx context.Context, item PT) (finalErr error) {
 	select {
 	case <-p.done:
 		return xerrors.WithStackTrace(errClosedPool)
+	case <-ctx.Done():
+		return xerrors.WithStackTrace(ctx.Err())
 	default:
-		if !item.IsAlive() {
+		p.stats.InUse().Dec()
```
asmyasnikov, line 403: manage counters must be outside from putItem call

internal/pool/pool.go
adwski: Changed as you suggested and also removed `<-ctx.Done()` inside, since it doesn't do anything

internal/pool/defaults.go
adwski: I guess it may be ok to retry instantly (createItem will wait for create timeout anyway).

internal/pool/pool.go
adwski: In this approach Idle() counter is showing amount of available items in queue channel, and because of this they are changed right after channel operation in getItem/putItem. Since getItem/putItem abstract queue interaction for try() and may discard item (instead of returning it (in get) or putting back to queue (in put)), managing Idle() counter outside of put/get will result in wrong value (for example when discarded items are still counted as idle).

There's not harm in moving InUse() counter outside. (Did it in last commit)

To sum up, I think It is fine if in some cases Idle+InUse < Limit, since imo counters represent stats about existing items. If item is discarded it should not be represented in stats.

PR #1317, opened 2024-07-01
* Added `query.{Client,Result}.Stats()` method

asmyasnikov:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-


internal/query/stats/query.go
```diff
@@ -0,0 +1,115 @@
+package stats
+
+import (
+	"time"
+
+	"github.com/ydb-platform/ydb-go-genproto/protos/Ydb_TableStats"
+)
+
+type (
```
neyguvj, line 9: table service and query service use the same type for query stats
https://github.com/ydb-platform/ydb-api-protos/blob/master/protos/ydb_query.proto#L187
https://github.com/ydb-platform/ydb-api-protos/blob/master/protos/ydb_table.proto#L817
so, you can use common type for both services

neyguvj (changes requested):

PR #1406, opened 2024-08-12
Changed query.Client API

asmyasnikov:
* Splitted experimental method `query.Client.Execute` to methods `query.Client.Exec` without result and `query.Client.Query` with result
* Renamed experimental method `query.Client.ReadResultSet` to `query.Client.QueryResultSet`
* Renamed experimental method `query.Client.ReadRow` to `query.Client.QueryRow`
* Removed experimental methods `query.Session.ReadResultSet` and  `query.Session.ReadRows`
* Removed experimental methods `query.Transaction.ReadResultSet` and  `query.Transaction.ReadRows`
* Removed experimental method `query.Client.Stats`

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes)
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-


query/session.go
```diff
@@ -20,28 +20,19 @@ type (
...
 		SessionInfo
 
-		// Execute executes query.
+		// Exec executes query.
 		//
-		// Execute used by default:
+		// Exec used by default:
 		// - DefaultTxControl (NoTx)
-		// - flag WithKeepInCache(true) if params is not empty.
-		Execute(ctx context.Context, query string, opts ...options.ExecuteOption) (tx Transaction, r Result, err error)
```
pelageech, line 28: Isn't it radical to remove Execute()? Why not make it deprecated and call a new Query(...) func under the hood? 
I understand that this is an experimental feature, but it is the most frequently used since this is the base. It seems that in such places it is better not to break compatibility right away. 
What do you think?

asmyasnikov: Execute splitted to Exec and Query by next reasons:
1) required read Result to io.EOF because otherwise next calls Execute happens to SESSION_BUSY or happens the intermediate errors which will not processed - next Execute's will have BAD_SESSION or NOT_FOUND 
2) required to Close Result for free resources
Users can forget read result to io.EOF or call result.Close method
Thats why method Exec do inside the reading to io.EOF and processing an intermediate errors with final result.Close 
Method Query returns known Result type for iterate over result sets and rows

asmyasnikov: Also we are removed Transaction from result tuple for unification signatures for Client, Session and Transaction
Users can get explicit Transaction using session.Begin call
This method is `noop` inside and returns lazy transaction. Lazy transaction will fully initialized on first call Exec or Query

pelageech: OK, thx for the explanation. The methods are different, but the following breaking changes make things really better:)

PR #1422, opened 2024-08-20
operation service usage in tests and example

asmyasnikov:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

### DIFF
--- a/operation/example_test.go
+++ b/operation/example_test.go
@@ -29,7 +29,7 @@ func Example_listOperations() {
 		for _, op := range operations.Operations {
 			fmt.Println(" -", op)
 		}
-		if len(operations.Operations) == 0 || nextToken == "" {
+		if nextToken == "" {
 			break
 		}
 	}
--- a/tests/integration/operation_test.go
+++ b/tests/integration/operation_test.go
@@ -39,7 +39,7 @@ func TestOperationList(t *testing.T) {
 
 	var nextToken string
 	for {
-		operations, err := db.Operation().List(ctx, operation.KindBuildIndex,
+		operations, err := db.Operation().List(ctx, operation.KindImportS3,
 			operation.WithPageSize(10),
 			operation.WithPageToken(nextToken),
 		)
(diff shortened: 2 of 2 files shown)


pelageech:
What happens when I call `DoTx(ctx, tx)` with serializable TxSettings and then I call tx.Query(..., **query.WithTxControl(query.OnlineReadOnlyTxControl())**). What an isolation do I have here?

Also, I noticed that `DoTx` takes `OnlineReadOnly` setting now. So, I'm interested in behaviour of serializable `tx.Query` inside `DoTx` with `OnlineReadOnly`.

pelageech:
Just an example.
```go
package main

import (
	"context"
	"fmt"
	"log"
	"time"

	"github.com/ydb-platform/ydb-go-sdk/v3"
	"github.com/ydb-platform/ydb-go-sdk/v3/query"
)

func main() {
	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
	defer cancel()
	driver, err := ydb.Open(ctx, "grpc://localhost:2136/local")

	if err != nil {
		log.Fatal("open", err)
	}
	defer driver.Close(ctx)

	var i int64
	err = driver.Query().DoTx(ctx, func(ctx context.Context, s query.TxActor) error {
		row, err := s.QueryRow(ctx, `SELECT 42;`, query.WithTxControl(query.OnlineReadOnlyTxControl())) // reduce isolation?

		if err != nil {
			log.Fatal("query row", err)
		}

		return row.Scan(&i)
	}, query.WithTxSettings(query.TxSettings(query.WithSerializableReadWrite()))) // serializable
	if err != nil {
		log.Fatal("DoTx", err)
	}

	fmt.Printf("%d\n", i)
}
```
It works but what happens under the hood. Does it reduce an isolation. And if I swap isolation, should I expect that an isolation will be straightened?

PR #1514, opened 2024-10-17
for review only

rekby:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-


tests/integration/table_regression_test.go
```diff
@@ -258,41 +223,6 @@ SELECT CAST($val AS UUID)
...
-
-			err = res.Scan(&resultFromDb)
-			if err != nil {
-				return err
-			}
-
-			return nil
-		})
-
-		require.NoError(t, err)
```
qrort, line 293: Should we remove these tests or check for Error instead?

rekby: I removed some function and the code cause compile error.
I don't know how to test in CI :(

PR #1584, opened 2024-12-13
Pool for encoders

DenisEvd:
## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

Currently, a separate encoder is used for each message.

Issue Number: 1462

## What is the new behavior?

Using sync.Pool for encoders to load GC with temporary objects less. Now I have added a pool for all encoders that implements Reset(io.Writer)

## Other information
Before using sync.Pool:
image

After:
image

These diagrams show how previously most of the time was spent on garbage collection and object allocation, and now directly on compression


internal/topic/topicwriterinternal/message.go
```diff
@@ -157,27 +146,26 @@ func (m *messageWithDataContent) readDataToTargetCodec(codec rawtopiccommon.Code
...
-
 	reader := m.Data
 	if reader == nil {
 		reader = &bytes.Reader{}
 	}
-	bytesCount, err := io.Copy(encoder, reader)
-	if err == nil {
-		err = encoder.Close()
+
+	buf := &bytes.Buffer{}
```
rekby, line 154: What about use [xsync.Pool](https://pkg.go.dev/github.com/ydb-platform/ydb-go-sdk/v3/internal/xsync#Pool) - internal type wrapper over sync pool for get bytes buffer?

It is often operation - it will need for every message.

internal/topic/topicwriterinternal/encoders.go
```diff
@@ -31,22 +59,66 @@ func NewEncoderMap() *EncoderMap {
...
+	if err == nil {
+		err = enc.Close()
+	}
+	if err != nil {
+		return 0, err
+	}
+
+	resetableEnc, ok := enc.(PublicResetableWriter)
+	if ok {
+		e.mx.Lock()
```
rekby, line 87: Here you need two locks every time: for check pool and for put codec. It is difficult to read and to compute

What about create pool for every codec within AddEncoder func?
It allows to remove sync for the maps completely, because them will be changed within single thead on init only.

internal/topic/topicwriterinternal/encoders_test.go
```diff
@@ -206,3 +209,79 @@ func TestCompressMessages(t *testing.T) {
...
+		_, err := testEncoderMap.Encode(rawtopiccommon.CodecRaw, buf, []byte("test_data"))
+		require.NoError(t, err)
+		require.Len(t, testEncoderMap.p, 0)
+		require.Equal(t, "test_data", buf.String())
+	})
+
+	t.Run("ResetableWriterCustom", func(t *testing.T) {
+		testEncoderMap := NewEncoderMap()
+
+		customCodecCode := rawtopiccommon.Codec(5)
```
rekby, line 237: What about use code of gzip codec? Or use code from custom codec space (10000-20000)

CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+* Added the use of sync.Pool in EncoderMap and renamed it to MultiEncoder
```
rekby, line 1: Renaming in internal packages no need to be logger into public changelog. Most important: supported pool of encoders, which implement  ResetableWriter interface

rekby (commented):
Thanks for the PR, i have only few cosmetic suggestions

internal/topic/topicwriterinternal/encoders_test.go
rekby: Overwrite codec must work - it is one of expected scenarion: customer overwrite internal codec to own implementation (for example special settings of gzip of more efficient implementation, than stdlib).

DenisEvd: Should I add a test for codec overwriting case?

rekby (approved):
Thanks for the PR :)

PR #1599, opened 2024-12-27
support custom types in query result scanning

4el0ve4ek:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-


internal/value/cast.go
```diff
@@ -10,5 +17,28 @@ func CastTo(v Value, dst interface{}) error {
...
+
+		err := v.castTo(dv)
+		if err != nil {
+			return err
+		}
+
+		return scanner.Scan(*dv)
+	}
+
+	if scanner, has := dst.(Scanner); has {
```
rekby, line 35: I think special ydb scanner should be preferred over more common sql scanner.

internal/value/cast.go
```diff
@@ -10,5 +17,28 @@ func CastTo(v Value, dst interface{}) error {
...
+		return scanner.UnmarshalYDBValue(v)
+	}
+
 	return v.castTo(dst)
 }
+
+type Scanner interface {
+	UnmarshalYDBValue(value Value) error
```
rekby, line 43: YDB has [scanner interface](https://github.com/ydb-platform/ydb-go-sdk/blob/e241a4db80b486b6541ff1775aa0729e91df2bf8/internal/scanner/scanner.go#L149-L149)
already. What about to reuse it?

rekby (changes requested):
Hello, @4el0ve4ek, thanks for your pr.

I have two questions to the pr.

internal/value/cast.go
4el0ve4ek: Thats very cool interface to use! But now there is only [one](https://github.com/ydb-platform/ydb-go-sdk/blob/e241a4db80b486b6541ff1775aa0729e91df2bf8/internal/table/scanner/scan_raw.go#L22) implementation 
 and  i don't find easy way to convert  value.Value into scanner.RawValue. So i took an easy way and create new interface)
 
 How do you think, maybe better to create adapter for value.Value which implements scanner.RawValue?

internal/value/cast.go
```diff
@@ -10,5 +17,20 @@ func CastTo(v Value, dst interface{}) error {
 		return nil
 	}
 
+	if _, ok := dst.(*uuid.UUID); ok {
```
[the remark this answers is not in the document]
4el0ve4ek: I made this crutch, because uuid.UUID [implements](https://github.com/google/uuid/blob/master/sql.go#L15) sql.Scanner, but in  [here](https://github.com/ydb-platform/ydb-go-sdk/blob/master/internal/value/value.go#L2602) we store object with type uuid.UUID as driver.Value, which is unexpected due to [go-docs](https://pkg.go.dev/database/sql/driver#Value). 

If you have suggestions how to fix it, i can make them)

internal/value/cast_test.go
```diff
@@ -428,6 +442,13 @@ func TestCastTo(t *testing.T) {
 			exp:   DateValueFromTime(time.Date(2024, 1, 2, 0, 0, 0, 0, time.UTC)),
 			err:   nil,
 		},
+		{
+			name:  xtest.CurrentFileLine(),
+			value: TextValue("text-string"),
+			dst:   ptr[testStringSQLScanner](),
+			exp:   testStringSQLScanner("text-string"),
+			err:   nil,
+		},
```
rekby, line 451:
```suggestion
		{
			name:  xtest.CurrentFileLine(),
			value: TextValue("text-string"),
			dst:   ptr[testStringSQLScanner](),
			exp:   testStringSQLScanner("text-string"),
			err:   nil,
		},
		{
			name:  xtest.CurrentFileLine(),
			value: Uuid(uuid.MustParse("6E73B41C-4EDE-4D08-9CFB-B7462D9E498B")),
			dst:   ptr[uuid.UUID](),
			exp:   uuid.MustParse("6E73B41C-4EDE-4D08-9CFB-B7462D9E498B"),
			err:   nil,
		},		
```

rekby (approved):
@asmyasnikov 
I have checked uuid, check the PR from your point of view about global cast logic please

PR #1603, opened 2025-01-09
fix broken metric `ydb_go_sdk_ydb_database_sql_conns`

qrort:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Counter `ydb_go_sdk_ydb_database_sql_conns` goes in negative values when `database/sql` driver is used. The reason is [after this commit](https://github.com/ydb-platform/ydb-go-sdk/commit/fd4142df6e70e949cc2b141bdc358d27473f69f9#diff-bc3f866611d992e908970b6cc021a96cb80ac60cbe0afd9372284c9713add5c3R51), `OnConnBeginTx` is called on begin, `OnConnClose` called on close, gauge is only decremented, and `database/sql` driver in sdk does not have its own `OnConnBeginTx` implementation, so it does nothing
Issue Number: N/A

## What is the new behavior?

Restores the behaviour before the mentioned commit

### DIFF
--- a/internal/xsql/conn.go
+++ b/internal/xsql/conn.go
@@ -48,7 +48,7 @@ func (c *Conn) CheckNamedValue(value *driver.NamedValue) (finalErr error) {
 }
 
 func (c *Conn) BeginTx(ctx context.Context, opts driver.TxOptions) (_ driver.Tx, finalErr error) {
-	onDone := trace.DatabaseSQLOnConnBeginTx(c.connector.trace, &ctx,
+	onDone := trace.DatabaseSQLOnConnBegin(c.connector.trace, &ctx,
 		stack.FunctionID("", stack.Package("database/sql")),
 	)
 	defer func() {
--- a/internal/xsql/conn.go
+++ b/internal/xsql/conn.go
@@ -48,7 +48,7 @@ func (c *Conn) CheckNamedValue(value *driver.NamedValue) (finalErr error) {
 }
 
 func (c *Conn) BeginTx(ctx context.Context, opts driver.TxOptions) (_ driver.Tx, finalErr error) {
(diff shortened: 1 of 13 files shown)


internal/xsql/conn.go
```diff
@@ -48,7 +48,7 @@ func (c *Conn) CheckNamedValue(value *driver.NamedValue) (finalErr error) {
 }
 
 func (c *Conn) BeginTx(ctx context.Context, opts driver.TxOptions) (_ driver.Tx, finalErr error) {
-	onDone := trace.DatabaseSQLOnConnBeginTx(c.connector.trace, &ctx,
+	onDone := trace.DatabaseSQLOnConnBegin(c.connector.trace, &ctx,
```
asmyasnikov, line 51: лучше в metrics/sql.go сделать фикс типа такого
```go
	t.OnConnBeginTx = func(info trace.DatabaseSQLConnBeginTxStartInfo) func(trace.DatabaseSQLConnBeginTxDoneInfo) {
		start := time.Now()
		if config.Details()&trace.DatabaseSQLTxEvents != 0 {
			return func(info trace.DatabaseSQLConnBeginTxDoneInfo) {
				txBegin.With(map[string]string{
					"status": errorBrief(info.Error),
				}).Inc()
				txBeginLatency.With(nil).Record(time.Since(start))
			}
		}

		return nil
	}
```

asmyasnikov: я специально разделил эти 2 колбэка чтобы считать/мониторить/трейсить вызовы Begin и BeginTx по-отдельности

asmyasnikov: этот фикс все равно нужен, но кажется он не поможет правильно считать метрику ydb_go_sdk_ydb_database_sql_conns

PR #1642, opened 2025-02-10
add WithReaderLogContext, WithWriterLogContext options to Topic API

qrort:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Logs written by background topic operations launched by `StartWriter`, `StartReader`, etc. can not be structured by client code in any way.

Issue Number: N/A

## What is the new behavior?

This PR adds new reader and writer options, `WithReaderLogContext` and `WithWriterLogContext`, which can be used to provide a `context.Context` instance to be used in logs.


qrort:
draft for now, updating and writing tests

internal/topic/topicclientinternal/client.go
```diff
@@ -356,15 +356,16 @@ func (c *Client) createWriterConfig(
 	topicPath string,
 	opts []topicoptions.WriterOption,
 ) topicwriterinternal.WriterReconnectorConfig {
-	var connector topicwriterinternal.ConnectFunc = func(ctx context.Context, tracer *trace.Topic) (
-		topicwriterinternal.RawTopicWriterStream,
-		error,
-	) {
-		return c.rawClient.StreamWrite(ctx, tracer)
-	}
+	//var connector topicwriterinternal.ConnectFunc = func(ctx context.Context, tracer *trace.Topic) (
```
rekby, line 359: remove the code instead of comment please

internal/topic/topicreadercommon/committer.go
```diff
@@ -148,6 +148,7 @@ func (c *Committer) pushCommitsLoop(ctx context.Context) {
 
 		onDone := trace.TopicOnReaderSendCommitMessage(
 			c.tracer,
+			&ctx,
```
rekby, line 151: Here and in other place. 
Don't transfer pointer to owned context directly - you shold copy local context, pass pointer to copy, them copy result to local context.

Pointer to context may be stored and then context can be changed by other code and data race may be happen.

[example](https://github.com/ydb-platform/ydb-go-sdk/blob/b035f5d08e59ef5480efcf31b9eae4171c253250/internal/topic/topicwriterinternal/writer_transaction.go#L34-L43)

internal/topic/topicreaderinternal/reader.go
```diff
@@ -68,11 +68,11 @@ func NewReader(
 	consumer string,
 	readSelectors []topicreadercommon.PublicReadSelector,
 	opts ...PublicReaderOption,
-) (Reader, error) {
+) (Reader, *context.Context, error) {
```
rekby, line 71: Why you want return context pointer?
It is very uncommon way.

rekby: What about add method to get current context to stream reader, reader reconnector, etc?
And it must not be context pointer - for prevent races while context changing.

rekby (changes requested):

topic/topicoptions/topicoptions_writer.go
```diff
@@ -214,3 +215,11 @@ func WithWriterTrace(t trace.Topic) WriterOption { //nolint:gocritic
 func WithWriterUpdateTokenInterval(interval time.Duration) WriterOption {
 	return topicwriterinternal.WithTokenUpdateInterval(interval)
 }
+
+// WithWriterLogContext allows providing a context.Context instance which will be used
+// in log/topic events.
+func WithWriterLogContext(ctx context.Context) WriterOption {
+	return func(cfg *topicwriterinternal.WriterReconnectorConfig) {
+		cfg.LogContext = ctx
```
rekby, line 223: I think internal names of log-context for reader and writer should be same.

trace/topic.go
```diff
@@ -153,9 +153,10 @@ type (
 
 	// Internals: https://github.com/ydb-platform/ydb-go-sdk/blob/master/VERSIONING.md#internals
 	TopicReaderStartInfo struct {
-		ReaderID int64
-		Consumer string
-		Error    error
+		LogContext *context.Context
```
rekby, line 156: What abount simple Context instead of LogContext?

qrort: I would prefer to keep it as-is to mark what the context is for

internal/topic/topicwriterinternal/writer_options.go
```diff
@@ -71,6 +72,14 @@ func WithCredentials(cred credentials.Credentials) PublicWriterOption {
 	}
 }
 
+// WithCredentials for internal usage only
```
rekby, line 75:
```suggestion
// WithRawClient for internal usage only
```

internal/topic/topicwriterinternal/writer_reconnector.go
```diff
@@ -112,6 +112,22 @@ func NewWriterReconnectorConfig(options ...PublicWriterOption) WriterReconnector
 		WithProducerID(uuid.NewString())(&cfg)
 	}
 
+	if cfg.Connect == nil {
```
rekby, line 115: The confgin should not store state and should not change self content - it is un-intuitive and difficult to understande when research the code.

internal/topic/topicreaderinternal/stream_reader_impl.go
```diff
@@ -185,21 +185,43 @@ func (r *topicStreamReaderImpl) WaitInit(_ context.Context) error {
...
-		tx.SessionID(),
-		tx,
-	)
-	ctx = traceCtx
+	onDone := func(err error) {
+		r.m.WithLock(func() {
+			logCtx := r.cfg.LogContext
+			trace.TopicOnReaderStreamPopBatchTx(
+				r.cfg.Trace,
+				&logCtx,
```
rekby, line 215: The code change behaviour: now is impossible get context from caller to tracer

internal/topic/topicreaderinternal/stream_reader_impl.go
```diff
@@ -185,21 +185,43 @@ func (r *topicStreamReaderImpl) WaitInit(_ context.Context) error {
...
+				r.readerID,
+				r.readConnectionID,
+				tx.SessionID(),
+				tx,
+			)(err)
+			r.cfg.LogContext = logCtx
```
rekby, line 221: 1. a config is not storage, the content of config should not changed after receive.
2. the context field rewrite in many placed without sync - it is good way for data races
3. The context infinite accumulate any additional fields from traces - it is good way for memory leak

rekby (changes requested):
Thanks for the PR and intresting for improve sdk.

I see few problems (write in code comments).

internal/xcontext/merge_contexts.go
```diff
@@ -0,0 +1,45 @@
+package xcontext
+
+import (
+	"context"
+	"time"
+)
+
+type MergedContexts struct {
+	context.Context                 //nolint:containedctx
```
rekby, line 9: Do you need context.Context field? I don't see usage of the field.

What about rename:
RequestContext -> DeadlineContext
BaseContext -> AdditionalValues

and make them private

internal/xcontext/merge_contexts.go
```diff
@@ -0,0 +1,45 @@
...
+
+func (ctx MergedContexts) Value(key interface{}) interface{} {
+	if ctx.RequestContext.Value(key) != nil {
+		return ctx.RequestContext.Value(key)
+	}
+
+	return ctx.BaseContext.Value(key)
+}
+
+func (ctx MergedContexts) AsCtx() *context.Context {
```
rekby, line 34: Do you really need the method?
may be better some helper function for use with any context?

some similar to:
```
ctxPointer(ctx *context.Context, f func (ctx context.Context)) {
  copyCtx := *ctx
  f(&copyCtx)
  *ctx = copyCtx
})
```
?

tests/integration/topic_log_test.go
```diff
@@ -0,0 +1,267 @@
...
+		return nil
+	})
+	require.NoError(t, err)
+
+	sendCDCMessage(ctx, t, db)
+	batch, err := reader.ReadMessagesBatch(ctx)
+	require.NoError(t, err)
+	require.NotEmpty(t, batch.Messages)
+
+	fileData, err := os.ReadFile("topic_log.json")
```
rekby, line 258: We don't want to guarantee stable text of logs.

I think it is bad idea to canonization of the logs.

internal/grpcwrapper/rawtopic/client.go
```diff
@@ -104,6 +104,7 @@ func (c *Client) StreamRead(ctxStreamLifeTime context.Context) (rawtopicreader.S
 func (c *Client) StreamWrite(
 	ctxStreamLifeTime context.Context,
 	tracer *trace.Topic,
+	logContext *context.Context,
```
rekby, line 107: What about merge contextx before  call StreamWrite?

why do you need a pointer?

qrort: merged before call, removed additional arg

internal/xcontext/merge_contexts.go
qrort: removed `AsCtx()`, helper function is not really needed because it creates more boilerplate when `trace.DoSmth(ctx, arg1, arg2)` is wrapped in `f` with one argument

internal/xcontext/merge_contexts.go
qrort: done

internal/topic/topicwriterinternal/encoders_test.go
```diff
@@ -15,14 +16,32 @@ import (
 	"github.com/ydb-platform/ydb-go-sdk/v3/trace"
 )
 
+func NewTestEncoderSelector(
```
rekby, line 19: I don't think it's necessary to create the helper function NewTestEncoderSelector. It's used only three times in the tests and only adds endpty context. I think it would be better to use EncoderSelector directly

internal/xcontext/merge_contexts.go
```diff
@@ -0,0 +1,40 @@
...
+
+func (ctx *MergedContexts) Value(key interface{}) interface{} {
+	if ctx.deadlineContext.Value(key) != nil {
+		return ctx.deadlineContext.Value(key)
+	}
+
+	return ctx.additionalValues.Value(key)
+}
+
+func MergeContexts(req context.Context, log context.Context) context.Context {
```
rekby, line 35: What about:
1. Renaming the arguments to `deadline` and `additionalValue/values`, similar to the internal field names?
2. Adding a comment like: "MergeContexts returns a context with the deadline from the `deadline` context and values from both contexts"?

log/context.go
```diff
@@ -37,5 +37,9 @@ func NamesFromContext(ctx context.Context) []string {
 }
 
 func with(ctx context.Context, lvl Level, names ...string) context.Context {
+	if ctx == nil {
```
rekby, line 40: Why do you need the check? Allowing a nil context usually leads to a panic if someone passes it. The standard library and others typically don't add guards for this situation. This helps prevent mistakes when someone forgets to pass a context, which can be hard to debug.

internal/topic/topicwriterinternal/encoders_test.go
qrort: removed NewTestEncoderSelector

rekby (approved):

PR #1650, opened 2025-02-17
Pool for decoders

dlc-01:
## Pull request type

- [ ] Bugfix
- [ *] Feature
- [ ] Build related changes

Issue Number: 1635

## What is the new behavior?

-  Using sync.Pool for decoders to load GC with temporary objects less.
-
-

### DIFF
--- a/internal/topic/topicreadercommon/decoders.go
+++ b/internal/topic/topicreadercommon/decoders.go
@@ -17,7 +17,7 @@ type decoderPool struct {
 
 func (p *decoderPool) Get() io.Reader {
 	dec, _ := p.pool.Get().(io.Reader)
-	
+
 	return dec
 }
 
@@ -26,7 +26,6 @@ func (p *decoderPool) Put(dec io.Reader) {
 }
 
 func newDecoderPool() *decoderPool {
-
 	return &decoderPool{
 		pool: sync.Pool{},
 	}
@@ -44,10 +43,12 @@ func NewDecoderMap() DecoderMap {
 	}
 
 	dm.AddDecoder(rawtopiccommon.CodecRaw, func(input io.Reader) (io.Reader, error) {
+
 		return input, nil
 	})
 
 	dm.AddDecoder(rawtopiccommon.CodecGzip, func(input io.Reader) (io.Reader, error) {
+
 		return gzip.NewReader(input)
 	})
 
@@ -66,9 +67,11 @@ func (m *DecoderMap) Decode(codec rawtopiccommon.Codec, input io.Reader) (io.Rea
 			var err error
 			decoder, err = f(input)
 			if err != nil {
+
 				return nil, err
 			}
 		}
+
 		return decoder, nil
 	}
 
--- a/internal/topic/topicreadercommon/decoders.go
+++ b/internal/topic/topicreadercommon/decoders.go
@@ -43,12 +43,10 @@ func NewDecoderMap() DecoderMap {
 	}
 
 	dm.AddDecoder(rawtopiccommon.CodecRaw, func(input io.Reader) (io.Reader, error) {
-
 		return input, nil
 	})
 
 	dm.AddDecoder(rawtopiccommon.CodecGzip, func(input io.Reader) (io.Reader, error) {
-
 		return gzip.NewReader(input)
 	})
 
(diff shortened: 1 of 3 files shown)


internal/topic/topicreadercommon/decoders.go
```diff
@@ -5,35 +5,71 @@ import (
...
 
 	"github.com/ydb-platform/ydb-go-sdk/v3/internal/grpcwrapper/rawtopic/rawtopiccommon"
 	"github.com/ydb-platform/ydb-go-sdk/v3/internal/xerrors"
 )
 
+type decoderPool struct {
+	pool sync.Pool
+}
+
+func (p *decoderPool) Get() io.Reader {
```
rekby, line 18: Decoder pool should return 
```go
type ReadResetter interface{
  io.Reader
  Reset(r io.Reader) error // - for switch decoder to new input
}
```

internal/topic/topicreadercommon/decoders.go
```diff
@@ -5,35 +5,71 @@ import (
...
+	dec, _ := p.pool.Get().(io.Reader)
+
+	return dec
+}
+
+func (p *decoderPool) Put(dec io.Reader) {
```
rekby, line 24: the method never called

internal/topic/topicreadercommon/decoders.go
```diff
@@ -5,35 +5,71 @@ import (
...
 func (m *DecoderMap) AddDecoder(codec rawtopiccommon.Codec, createFunc PublicCreateDecoderFunc) {
 	m.m[codec] = createFunc
+	m.dp[codec] = newDecoderPool()
 }
 
 func (m *DecoderMap) Decode(codec rawtopiccommon.Codec, input io.Reader) (io.Reader, error) {
 	if f := m.m[codec]; f != nil {
-		return f(input)
+		decoder := m.dp[codec].Get()
+		if decoder == nil {
```
rekby, line 64: in else brach no decoder switch to `input`

rekby (changes requested):
Hello, thanks for the PR. 
It is good start point, but it is not implement the task yet.

The PR no add usage of decoder pool

internal/topic/topicreadercommon/decoders.go
```diff
@@ -5,43 +5,122 @@ import (
...
+		return gz, nil
+	})
+
+	return dm
 }
 
-func (m *DecoderMap) AddDecoder(codec rawtopiccommon.Codec, createFunc PublicCreateDecoderFunc) {
+func (m *DecoderMap) AddDecoder(codec rawtopiccommon.Codec, createFunc func(io.Reader) (ReadResetter, error)) {
 	m.m[codec] = createFunc
+	m.dp[codec] = newDecoderPool()
```
rekby, line 68: What about support custom codec?

We must allow to accept custom external codec and save current interface.
For test you can create custom codec with read bytes in reverse order.

We should work with items, created with createFunc. Pool must be usable for custom codec too (for example -check implementation of ReadResetter before store item to pool).

rekby (changes requested):
Hello, thanks for your PR.

The code contains a bug: it lost available of use custom decoders (see comment).

Add a test for the case please.

rekby (changes requested):
Thanks for the PR, I have two queestion for it (in comments)

internal/topic/topicreadercommon/decoders.go
```diff
@@ -5,43 +5,122 @@ import (
...
 }
 
-func (m *DecoderMap) AddDecoder(codec rawtopiccommon.Codec, createFunc PublicCreateDecoderFunc) {
-	m.m[codec] = createFunc
+type pooledDecoder struct {
+	ReadResetter
+	pool *decoderPool
+}
+
+func (p *pooledDecoder) Close() error {
```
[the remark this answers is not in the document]
dlc-01: Since pooledDecoder is a wrapper over the decoder (ReadResetter) that adds a Close method.
Close must be called by the user of the Decode method, that is, the code that io.Reader got from m.Decode. e.g. 
``` 
reader, err := decoderMap.Decode(codec, input)
if err != nil {
    // Process the error
}
// Read data
data, err := io.ReadAll(reader)
if err != nil {
    // Process the error
}
// Close reader to return the decoder to the pool
if closer, ok := reader.(io.Closer); ok {
    err := closer.Close()
    if err != nil {
        // Process the error
    }
}
``` 

I'm going to analyze the code to see where it should be done.

PR #1675, opened 2025-03-07
Add ReadRows to the table.Client

robdrynkin:
## Pull request type

- [ ] Feature

Issue Number: N/A

## What is the new behavior?

Add ReadRows method to the table.Client, because it doesn't really use session, but sessions sometimes override PreferredNodeID.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-* Added ReadRows method for table.Client
+* Added `table.Client.ReadRows` method with internal retries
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+* Added ReadRows method for table.Client
```
asmyasnikov, line 1:
```suggestion
* Added `table.Client.ReadRows` method with internal retries
```

PR #1694, opened 2025-03-14
close dropped connection on balancer.Update()

qrort:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: 1692

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+* Pinged new connections on discovery attempt, closed dropped ones, so `ydb_go_sdk_ydb_driver_conns` metric is correct
+
 ## v3.111.0
 * Added `sugar.PrintErrorWithoutStack` helper for remove stack records from error string
 * Added `sugar.UnwrapError` helper for unwrap source error to root errors
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-* Pinged new connections on discovery attempt, closed dropped ones, so `ydb_go_sdk_ydb_driver_conns` metric is correct
+* Fixed `ydb_go_sdk_ydb_driver_conns` metric to correctly display nodes that are discovered at the moment
 
 ## v3.111.0
 * Added `sugar.PrintErrorWithoutStack` helper for remove stack records from error string
--- a/internal/balancer/balancer.go
+++ b/internal/balancer/balancer.go
@@ -175,23 +175,24 @@ func (b *Balancer) clusterDiscoveryAttempt(ctx context.Context, cc *grpc.ClientC
 	return nil
 }
 
-func buildConnectionsState(ctx context.Context, pool interface {
-	GetIfPresent(endpoint endpoint.Endpoint) conn.Conn
-	Allow(ctx context.Context, cc conn.Conn)
-	EndpointsToConnections(endpoints []endpoint.Endpoint) []conn.Conn
-}, newest []endpoint.Endpoint,
+func buildConnectionsState(ctx context.Context,
+	pool interface {
+		GetIfPresent(endpoint endpoint.Endpoint) conn.Conn
+		Allow(ctx context.Context, cc conn.Conn)
+		EndpointsToConnections(endpoints []endpoint.Endpoint) []conn.Conn
+	},
+	newest []endpoint.Endpoint,
 	dropped []endpoint.Endpoint,
-	config balancerConfig.Config,
+	balancerConfig *balancerConfig.Config,
 	selfLocation balancerConfig.Info,
 ) *connectionsState {
 	connections := pool.EndpointsToConnections(newest)
 	for _, c := range connections {
 		pool.Allow(ctx, c)
 		c.Endpoint().Touch()
-		_ = c.Ping(ctx)
 	}
 
-	state := newConnectionsState(connections, config.Filter, selfLocation, config.AllowFallback)
+	state := newConnectionsState(connections, balancerConfig.Filter, selfLocation, balancerConfig.AllowFallback)
 
 	for _, e := range dropped {
 		c := pool.GetIfPresent(e)
@@ -229,7 +230,7 @@ func (b *Balancer) applyDiscoveredEndpoints(ctx context.Context, newest []endpoi
 	}()
 
 	info := balancerConfig.Info{SelfLocation: localDC}
-	b.connectionsState.Store(buildConnectionsState(ctx, b.pool, newest, dropped, b.balancerConfig, info))
+	b.connectionsState.Store(buildConnectionsState(ctx, b.pool, newest, dropped, &b.balancerConfig, info))
 }
 
 func (b *Balancer) Close(ctx context.Context) (err error) {
--- a/internal/balancer/balancer.go
+++ b/internal/balancer/balancer.go
@@ -175,6 +175,34 @@ func (b *Balancer) clusterDiscoveryAttempt(ctx context.Context, cc *grpc.ClientC
 	return nil
 }
 
+func buildConnectionsState(ctx context.Context, pool interface {
+	GetIfPresent(endpoint endpoint.Endpoint) conn.Conn
+	Allow(ctx context.Context, cc conn.Conn)
+	EndpointsToConnections(endpoints []endpoint.Endpoint) []conn.Conn
+}, newest []endpoint.Endpoint,
+	dropped []endpoint.Endpoint,
+	config balancerConfig.Config,
+	selfLocation balancerConfig.Info,
+) *connectionsState {
+	connections := pool.EndpointsToConnections(newest)
+	for _, c := range connections {
+		pool.Allow(ctx, c)
+		c.Endpoint().Touch()
+		_ = c.Ping(ctx)
+	}
+
+	state := newConnectionsState(connections, config.Filter, selfLocation, config.AllowFallback)
+
+	for _, e := range dropped {
+		c := pool.GetIfPresent(e)
+		if c != nil {
+			_ = c.Close(ctx)
+		}
+	}
+
+	return state
+}
+
 func (b *Balancer) applyDiscoveredEndpoints(ctx context.Context, newest []endpoint.Endpoint, localDC string) {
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+* Ping new connections on discovery attempt, close dropped ones, so `ydb_go_sdk_ydb_driver_conns` metric is correct
```
asmyasnikov, line 1: глаголы в прошедшем времени

internal/mock/conn.go
```diff
@@ -116,7 +120,7 @@ func (e *Endpoint) LoadFactor() float32 {
 }
 
 func (e *Endpoint) OverrideHost() string {
-	panic("not implemented in mock")
+	return ""
```
asmyasnikov, line 123: а это зачем потрогал? где-то вылезло при тестировании?

qrort: ага, теперь applyDiscoveredEndpoints зовет dial -> зовет OverrideHost

internal/balancer/balancer.go
```diff
@@ -198,21 +228,27 @@ func (b *Balancer) applyDiscoveredEndpoints(ctx context.Context, newest []endpoi
...
+	//state := newConnectionsState(connections, b.balancerConfig.Filter, info, b.balancerConfig.AllowFallback)
+	//
+	//b.connectionsState.Store(state)
+	//
+	//for _, e := range dropped {
+	//	c := b.pool.GetIfPresent(e)
+	//	if c != nil {
+	//		_ = c.Close(ctx)
+	//	}
+	//}
```
asmyasnikov, line 251:
```suggestion
```

internal/mock/conn.go
```diff
@@ -17,6 +17,8 @@ type Conn struct {
 	NodeIDField   uint32
 	State         conn.State
 	LocalDCField  bool
+	Pinged        bool
+	Closed        bool
```
asmyasnikov, line 21: может atomic.Bool?

internal/mock/conn.go
```diff
@@ -53,7 +55,19 @@ func (c *Conn) Park(ctx context.Context) (err error) {
 	panic("not implemented in mock")
 }
 
+func (c *Conn) Close(ctx context.Context) error {
+	c.Closed = true
```
asmyasnikov, line 59: строчкой ниже concurent safe вызов, а тут DATA RACE

internal/mock/conn.go
```diff
@@ -53,7 +55,19 @@ func (c *Conn) Park(ctx context.Context) (err error) {
...
+	c.SetState(ctx, conn.Offline)
+
+	return nil
+}
+
 func (c *Conn) Ping(ctx context.Context) error {
+	c.Pinged = true
```
asmyasnikov, line 66: аналогично

asmyasnikov (approved):

internal/balancer/balancer.go
```diff
@@ -175,6 +175,34 @@ func (b *Balancer) clusterDiscoveryAttempt(ctx context.Context, cc *grpc.ClientC
...
+func buildConnectionsState(ctx context.Context, pool interface {
+	GetIfPresent(endpoint endpoint.Endpoint) conn.Conn
+	Allow(ctx context.Context, cc conn.Conn)
+	EndpointsToConnections(endpoints []endpoint.Endpoint) []conn.Conn
+}, newest []endpoint.Endpoint,
+	dropped []endpoint.Endpoint,
+	config balancerConfig.Config,
+	selfLocation balancerConfig.Info,
+) *connectionsState {
+	connections := pool.EndpointsToConnections(newest)
```
rekby, line 187: Do I understand correctly that it will create a real connection for all endpoints here?

If so, it might be a problem for large clusters.

qrort: added a new trace.Driver hook, connection is not pinged on discovery now

metrics/driver.go
```diff
@@ -139,19 +139,15 @@ func driver(config Config) (t trace.Driver) {
...
-		return func(info trace.DriverConnDialDoneInfo) {
-			if config.Details()&trace.DriverConnEvents != 0 {
-				if info.Error == nil {
-					conns.With(map[string]string{
-						"endpoint": endpoint,
-						"node_id":  idToString(nodeID),
-					}).Add(1)
-				}
-			}
+		if config.Details()&trace.DriverConnEvents != 0 {
```
asmyasnikov: DriverBalancerEvents or DiscoveryEvents

metrics/driver.go
```diff
@@ -139,19 +139,15 @@ func driver(config Config) (t trace.Driver) {
 			}
 		}
 	}
-	t.OnConnDial = func(info trace.DriverConnDialStartInfo) func(trace.DriverConnDialDoneInfo) {
+	t.OnConnDiscover = func(info trace.DriverConnDiscoverInfo) {
```
asmyasnikov: why you drop OnConnDial callabck?

trace/driver.go
```diff
@@ -52,6 +52,8 @@ type (
 		// Internals: https://github.com/ydb-platform/ydb-go-sdk/blob/master/VERSIONING.md#internals
 		OnConnDial func(DriverConnDialStartInfo) func(DriverConnDialDoneInfo)
 		// Internals: https://github.com/ydb-platform/ydb-go-sdk/blob/master/VERSIONING.md#internals
+		OnConnDiscover func(DriverConnDiscoverInfo)
```
asmyasnikov: OnBalancerDiscover or OnDriverDiscover

metrics/driver.go
qrort: OnConnDial does the same in terms of metics, and now that I increase it on discover, i do not need a Dial hook

trace/driver.go
```diff
@@ -282,10 +294,24 @@ type (
...
+		// Context make available context in trace callback function.
+		// Pointer to context provide replacement of context in trace callback function.
+		// Warning: concurrent access to pointer on client side must be excluded.
+		// Safe replacement of context are provided only inside callback function
+		Context  *context.Context
+		Call     call
+		Endpoint EndpointInfo
+	}
+	// Internals: https://github.com/ydb-platform/ydb-go-sdk/blob/master/VERSIONING.md#internals
+	DriverConnDiscoverDoneInfo struct {
```
asmyasnikov: Conn - лишнее

kprokopenko (approved):

metrics/driver.go
```diff
@@ -139,19 +139,15 @@ func driver(config Config) (t trace.Driver) {
 			}
 		}
 	}
-	t.OnConnDial = func(info trace.DriverConnDialStartInfo) func(trace.DriverConnDialDoneInfo) {
+	t.OnBalancerDiscover = func(info trace.DriverBalancerDiscoverInfo) {
```
kprokopenko: Found at the last moment. Why is the traces changing? Is something said about this in the task? Or is there no other way to solve the initial problem?

log/driver.go
```diff
@@ -460,7 +460,7 @@ func internalDriver(l Logger, d trace.Detailer) trace.Driver {
 			if d.Details()&trace.DriverBalancerEvents == 0 {
 				return nil
 			}
-			ctx := with(*info.Context, TRACE, "ydb", "driver", "balancer", "update")
+			ctx := with(*info.Context, INFO, "ydb", "driver", "balancer", "update")
```
kprokopenko, line 463: Is this a coordinated (requested by somebody) change in the level?

internal/balancer/balancer.go
```diff
@@ -186,10 +215,12 @@ func (b *Balancer) applyDiscoveredEndpoints(ctx context.Context, newest []endpoi
 		)
 		previous = b.connections().All()
 	)
+
```
kprokopenko, line 218: Now the values for the `added` and `dropped` metrics are calculated before closing connections in the `buildConnectionsState` method. I just want to draw attention to this.

metrics/driver.go
qrort: Before this commit:
1. Metric increments on conn dial, which can happen way after the discovery, and clients do not see valid set of nodes that SDK can connect to.
2. Metric decrements on conn close, which is sometimes not called, because the discovery event changes connection state without calling `Close()`.
After this commit:
1. Metric increments on connection discovery, which solves the problem above. after this change, `OnConnDial()` hook for metrics becomes empty, that is why it is removed from `t`.
2. `Close()` is now called for dropped connections.

qrort: you can suggest other approach, but i do not see what is wrong with this one

log/driver.go
qrort: It is requested by the client, and seems logical, because discovery events are rare, do not pollute the logs and show meaningful info.

internal/balancer/balancer.go
qrort: `added` and `dropped` represent the discovery call diff, as it was before the commit. We can add "DroppedAndClosed" array and pass it to logs, do you think it is needed?

kprokopenko: No, thanks, and thank you for the explanation

metrics/driver.go
kprokopenko: The idea is clear, but I'm concerned that the metrics will change for existing users. I need some time to discuss it.

internal/balancer/balancer.go
```diff
@@ -175,6 +175,35 @@ func (b *Balancer) clusterDiscoveryAttempt(ctx context.Context, cc *grpc.ClientC
...
+		pool.Allow(ctx, c)
+		c.Endpoint().Touch()
+	}
+
+	state := newConnectionsState(connections, balancerConfig.Filter, selfLocation, balancerConfig.AllowFallback)
+
+	for _, e := range dropped {
+		c := pool.GetIfPresent(e)
+		if c != nil {
+			_ = c.Close(ctx)
```
kprokopenko, line 200: When the connection is closed, requests that are performing long-running stream reading may be interrupted. Required graceful termination of requests.

Issue #1511, opened 2024-10-17
feat: Split bulk for bulkupsert


rekby:
Split bulk of `BulkUpsert` if the bulk greater then max grpc message size:
```
transport/ResourceExhausted (code = 8, source error = \"rpc error: code = ResourceExhausted desc = trying to send message larger than max (75139668 vs. 67108864)\"
```

as first step - doesn't retry the error.

rekby:
it will have problem with non protobuf formats, for example arrow and csv.

In the ticket will fix bulk upsert with native sdk types only.

PR #1803, opened 2025-06-06
Split bulk for bulkupsert

kprokopenko:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

ResourceExhausted if the bulk greater than max GRPC request size (64M).

Issue Number: #1511

## What is the new behavior?

If a bulk request exceeds the maximum size allowed by gRPC, it is split into smaller chunks so that each chunk does not exceed the limit. Each such request is sent separately.

-
-
-


driver.go
```diff
@@ -470,6 +470,9 @@ func (d *Driver) connect(ctx context.Context) (err error) {
 					// prepend common params from root config
 					[]tableConfig.Option{
 						tableConfig.With(d.config.Common),
+
+						// because we are limiting by grpc message size
```
rekby: I think it is not need additional comments

internal/table/client.go
```diff
@@ -319,22 +321,88 @@ func (c *Client) BulkUpsert(
...
+	}
+
+	return append(leftChunks, rightChunks...), nil
+}
+
+// splitBulkUpsertRequestAt splits a BulkUpsertRequest into two parts at the specified position.
+// It creates a new request with items from the specified position to the end,
+// while modifying the original request to contain only items before the position.
+// Returns the newly created request with the split items.
+func splitBulkUpsertRequestAt(req *Ydb_Table.BulkUpsertRequest, pos int) *Ydb_Table.BulkUpsertRequest {
```
rekby: What about return both requests: left and right? I think it will be simpler to understand.

It is uncommon pattern: cut part of input item and  return them separately. May be rename the func.
From split I wait many items in results (for this case two items instead of a slice), similar to https://pkg.go.dev/strings#Split,

tests/integration/table_bulk_upsert_test.go
```diff
@@ -243,3 +273,12 @@ func assertIdValue(ctx context.Context, t *testing.T, tableName string, id int64
 func assertIdValueNil(ctx context.Context, t *testing.T, tableName string, id int64) {
 	assertIdValueImpl(ctx, t, tableName, id, nil)
 }
+
+func stringWithSize(sizeBytes int) string {
```
rekby: what about https://pkg.go.dev/strings#Repeat instead of this function?

internal/table/client_test.go
```diff
@@ -80,6 +83,56 @@ func TestRaceWgClosed(t *testing.T) {
...
+		require.NoError(t, err)
+		assert.Len(t, got, 2)
+		assert.Less(t, proto.Size(got[0]), 100)
+		assert.Less(t, proto.Size(got[1]), 100)
+	})
+
+	t.Run("zero max size", func(t *testing.T) {
+		input := newTestBulkRequest(t, 50)
+		_, err := chunkBulkUpsertRequest(input, 0)
+		assert.ErrorContains(t, err, "ydb: single row size (27 bytes) exceeds maximum request size (0 bytes) - row is too large to process")
```
rekby: We doesn't test error strings and don't provide errors text stable guarantees

internal/table/client.go
```diff
@@ -319,22 +321,88 @@ func (c *Client) BulkUpsert(
...
+		}
+	}
+
 	return nil
 }
 
+// chunkBulkUpsertRequest splits a BulkUpsertRequest into smaller chunks if it exceeds the maximum message size.
+// It recursively divides the request's rows into smaller requests while preserving the original request structure.
+// If the request is smaller than the maximum size or contains fewer than two rows, it returns the original request.
+func chunkBulkUpsertRequest(req *Ydb_Table.BulkUpsertRequest, maxBytes int) ([]*Ydb_Table.BulkUpsertRequest, error) {
```
rekby: what about accept `dst []*Ydb_Table.BulkUpsertRequest` and always append new items to dst, for prevent allocate new slice for every iteration.

internal/table/client.go
```diff
@@ -348,59 +356,60 @@ func (c *Client) BulkUpsert(
...
 // It recursively divides the request's rows into smaller requests while preserving the original request structure.
 // If the request is smaller than the maximum size or contains fewer than two rows, it returns the original request.
-func chunkBulkUpsertRequest(req *Ydb_Table.BulkUpsertRequest, maxBytes int) ([]*Ydb_Table.BulkUpsertRequest, error) {
-	// not a row bulk upsert request -> ret original request
-	if req.Rows == nil || req.Rows.Value == nil {
-		return []*Ydb_Table.BulkUpsertRequest{req}, nil
-	}
-
+func chunkBulkUpsertRequest(
+	dst *[]*Ydb_Table.BulkUpsertRequest,
```
rekby: Pointer to slice is uncommon pattern, what about return slice with appended requests, similar to builtin `append`?

internal/table/client.go
```diff
@@ -348,59 +356,60 @@ func (c *Client) BulkUpsert(
...
 	if reqSize <= maxBytes {
-		return []*Ydb_Table.BulkUpsertRequest{req}, nil
+		*dst = append(*dst, req)
+
+		return nil
+	}
+
+	// not a row bulk upsert request -> ret original request
+	if req.GetRows() == nil || req.GetRows().GetValue() == nil {
+		return fmt.Errorf("ydb: request size (%d bytes) exceeds maximum size (%d bytes) "+
```
rekby: add xerrors.WithStacktrace(xerrors.Wrap(...))

internal/table/client.go
```diff
@@ -348,59 +356,60 @@ func (c *Client) BulkUpsert(
...
 	if n == 0 {
-		return []*Ydb_Table.BulkUpsertRequest{req}, nil
+		return nil
 	}
 
 	// we cannot split one item and one item is too big
 	if n == 1 {
-		err := fmt.Errorf("ydb: single row size (%d bytes) exceeds maximum request size (%d bytes) - row is too large to process", reqSize, maxBytes)
-		return nil, err
+		return fmt.Errorf("ydb: single row size (%d bytes) exceeds maximum request size (%d bytes) "+
```
rekby: need to wrap this error

internal/table/client.go
```diff
@@ -348,59 +356,60 @@ func (c *Client) BulkUpsert(
...
 // while modifying the original request to contain only items before the position.
 // Returns the newly created request with the split items.
-func splitBulkUpsertRequestAt(req *Ydb_Table.BulkUpsertRequest, pos int) *Ydb_Table.BulkUpsertRequest {
-	items := req.Rows.Value.Items // save original items
+func splitBulkUpsertRequestAt(req *Ydb_Table.BulkUpsertRequest, pos int) (_, _ *Ydb_Table.BulkUpsertRequest) {
+	items := req.GetRows().GetValue().GetItems() // save original items
 	req.Rows.Value.Items = nil
 
-	right := proto.Clone(req).(*Ydb_Table.BulkUpsertRequest)
+	right, _ := proto.Clone(req).(*Ydb_Table.BulkUpsertRequest)
```
rekby: Is it error ignore? when the error can happened?

What about return error or panic (if error not possible in real case)?

internal/table/client.go
```diff
@@ -295,13 +295,33 @@ func (c *Client) BulkUpsert(
...
+	opts ...table.Option,
+) (finalErr error) {
 	attempts, config := 0, c.retryOptions(opts...)
 	config.RetryOptions = append(config.RetryOptions,
 		retry.WithIdempotent(true),
 		retry.WithTrace(&trace.Retry{
 			OnRetry: func(info trace.RetryLoopStartInfo) func(trace.RetryLoopDoneInfo) {
 				return func(info trace.RetryLoopDoneInfo) {
-					attempts = info.Attempts
+					attempts += info.Attempts - 1
```
rekby: What about chek if info.Attempts > 0?
If it will be zero - now we will decrement attempts

PR #1806, opened 2025-06-10
set session status when `closeOnce()` is called instead of `Close()`

qrort:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

session status can be wrong, and users can sometimes experience timout when using `ExecuteDataQueryOverQueryService`

Issue Number: N/A

## What is the new behavior?

fix resolves a concrete example of degradation on https://github.com/yandex/temporal-over-ydb/

### DIFF
--- a/internal/query/session_core_test.go
+++ b/internal/query/session_core_test.go
@@ -1,6 +1,11 @@
 package query
 
 import (
+	"context"
+	"errors"
+	"runtime"
+	"runtime/debug"
+	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


internal/query/session_core_test.go
```diff
@@ -58,5 +64,126 @@ func TestSessionCoreCancelAttachOnDone(t *testing.T) {
 		core.closeOnce()
 		require.GreaterOrEqual(t, recvMsgCounter.Load(), uint32(2))
 		require.LessOrEqual(t, recvMsgCounter.Load(), uint32(3))
+		require.Equal(t, core.Status(), StatusClosed.String())
+	}, xtest.StopAfter(time.Second))
+}
+
+func TestSessionCoreAttachError(t *testing.T) {
```
kprokopenko, line 71: These tests pass on the master branch. Could you add a test that reproduces the problem, i.e., fails on the master branch?

PR #1817, opened 2025-06-21
fixed redundant defer in sessionCore.closeOnce

asmyasnikov:


internal/query/session_core.go
```diff
@@ -198,8 +198,8 @@ func (core *sessionCore) attach(ctx context.Context) (finalErr error) {
 	}
 
 	core.closeOnce = sync.OnceFunc(func() {
-		defer close(core.done)
```
kprokopenko, line 202: Maybe we should keep `defer close()`?

kprokopenko (approved):

PR #1819, opened 2025-06-23
migrate golangci lint to v2

4el0ve4ek:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-


.golangci.yml
```diff
@@ -1,211 +1,15 @@
-# options for analysis running
+version: "2"
 run:
-  # exit code when at least one issue was found, default is 1
```
4el0ve4ek, line 3: migrated with golangci-lint migrate, and he dropped all comments
if they are important, i can try to recover them

.golangci.yml
```diff
@@ -238,76 +41,161 @@ linters:
     - varnamelen
     - wrapcheck
     - wsl
-
+    - recvcheck
+    - iface
+    - gosmopolitan
+    - funcorder
```
4el0ve4ek, line 47: new linters were added to the golangci-lint, and maintaining them was too much for this PR
recvcheck - Checks for receiver type consistency.
iface - Detect the incorrect use of interfaces, helping developers avoid interface pollution.
gosmopolitan - reports usage of time.Local, and it failed because time.Local used in tests
funcorder - Checks the order of functions, methods, and constructors.

Probably would be better to make different issues on supporting them? but except gosmopolitan, they require too big refactoring)

.golangci.yml
```diff
@@ -238,76 +41,161 @@ linters:
...
+          skipRecvDeref: false
+        unnamedResult:
+          checkExported: true
+    gomoddirectives:
+      replace-allow-list:
+        - xorm.io/xorm
+      replace-local: true
+    # govet:
+    #   enable:
+    #     - shadow #
```
4el0ve4ek: govet  wasn't working in previous iteration, and when i tried to enable it, he showed me places where err is shadowed ([here for example](https://github.com/ydb-platform/ydb-go-sdk/blob/master/internal/xsql/tx.go#L92-L98)). I didn't find a nice way to mute err case)

What do you think? Maybe fix it in other issues?

internal/credentials/static.go
```diff
@@ -75,7 +75,7 @@ type Static struct {
 	sourceInfo string
 }
 
-//nolint:funlen
+//nolint:funlen,nolintlint
```
kprokopenko:
```suggestion
```

You can safely remove nolint

internal/topic/topiclistenerinternal/stream_listener.go
```diff
@@ -187,7 +187,7 @@ func (l *streamListener) initVars(sessionIDCounter *atomic.Int64) {
 	}
 }
 
-//nolint:funlen
+//nolint:funlen,nolintlint
```
kprokopenko:
```suggestion
//nolint:funlen
```

No issues with `nolint` here

internal/scripting/client.go
```diff
@@ -162,7 +162,7 @@ func (c *Client) explain(ctx context.Context, sql string, mode scripting.Explain
 		result   = Ydb_Scripting.ExplainYqlResult{}
 	)
 	defer func() {
-		onDone(e.Explanation.Plan, err)
+		onDone(e.Plan, err)
```
kprokopenko: I don't see any issue from linter with `e.Explanation.Plan`.

internal/table/scanner/result.go
```diff
@@ -94,13 +94,13 @@ type option func(r *baseResult)
 
 func WithIgnoreTruncated(ignoreTruncated bool) option {
 	return func(r *baseResult) {
-		r.valueScanner.ignoreTruncated = ignoreTruncated
+		r.ignoreTruncated = ignoreTruncated
```
kprokopenko: No issue here. Let's return it.

.golangci.yml
kprokopenko: Not important

.golangci.yml
kprokopenko: Good point. I disabled `shadow`.

.golangci.yml
kprokopenko: Of course, new linters should be added in the separate PR.
I created issue https://github.com/ydb-platform/ydb-go-sdk/issues/1821

kprokopenko (approved):

PR #1840, opened 2025-07-29
Failover unit test for DNS with unavailable IP addresses

kprokopenko:
If there are issues with IP addresses via DNS, YDB startup gets stuck.

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): Test

The problem was fixed early, but no tests exist.

Issue Number: N/A

## What is the new behavior?

Unit test for guaranteed stability.

### DIFF
--- a/internal/balancer/balancer_test.go
+++ b/internal/balancer/balancer_test.go
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/require"
-
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/resolver"
 	"google.golang.org/grpc/test/bufconn"
@@ -41,7 +40,7 @@ func TestBalancer_discoveryConn(t *testing.T) {
 						atomic.AddUint32(&dialAttempt, 1)
 
 						if atomic.LoadUint32(&dialAttempt) == 1 {
-							time.Sleep(1 * time.Hour) //extremely slow dialing
+							time.Sleep(1 * time.Hour) // extremely slow dialing
 
 							return nil, fmt.Errorf("fake error for endpoint: %s", s)
 						}
--- a/internal/balancer/balancer_test.go
+++ b/internal/balancer/balancer_test.go
@@ -17,7 +17,22 @@ import (
 )
 
 func TestBalancer_discoveryConn(t *testing.T) {
+	// testTimeout defines the test timeout and is an example of an actual user-defined timeout.
+	//
+	// I couldn't find any events for synchronization, assuming that there
+	// might be retries with different logic and their own timeouts inside `discoveryConn`.
+	// If not now, then in the future. One second excludes false test failures in case
+	// the test is run on very slow workers. Moreover, one second is only lost if the test fails,
+	// and in that case, losing it is not critical. Upon successful completion of the test,
+	// the context is canceled via `cancel()`.
+	const testTimeout = 1 * time.Second
+
+	ctx, cancel := context.WithTimeout(context.Background(), testTimeout)
+	defer cancel()
+
 	fakeListener := bufconn.Listen(1024 * 1024)
+	defer fakeListener.Close()
+
 	fakeServer := grpc.NewServer()
 	defer fakeServer.Stop()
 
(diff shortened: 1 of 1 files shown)


internal/balancer/balancer_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/ydb-platform/ydb-go-sdk/v3/config"
+)
+
+func TestBalancer_discoveryConn(t *testing.T) {
+	fakeListener := bufconn.Listen(1024 * 1024)
+	fakeServer := grpc.NewServer()
+	defer fakeServer.Stop()
+
+	go func() {
+		_ = fakeServer.Serve(fakeListener)
```
rekby, line 40: Does the listener need to stop?

internal/balancer/balancer_test.go
```diff
@@ -0,0 +1,87 @@
...
+			config.WithGrpcOptions(
+				grpc.WithResolvers(&mockResolverBuilder{}),
+
+				grpc.WithContextDialer(
+					// The first dialing is very long and ends with an error, while the subsequent ones work fine.
+					func(ctx context.Context, s string) (net.Conn, error) {
+						atomic.AddUint32(&dialAttempt, 1)
+
+						if atomic.LoadUint32(&dialAttempt) == 1 {
+							time.Sleep(1 * time.Hour) // extremely slow dialing
```
rekby: I think about sleep a hour is bad think. For test it mean sleep forever and leak the goroutine.

What about use empty channel for signal about the function can be returned?
example: [stream_listener.go](https://github.com/ydb-platform/ydb-go-sdk/blob/d726bbdf94edf82a9234ee52881cbece4a3f5bd9/internal/topic/topiclistenerinternal/stream_listener.go#L194-L194)

internal/balancer/balancer_test.go
```diff
@@ -0,0 +1,87 @@
...
+					}),
+
+				// If you want to reproduce the issue, uncomment the line:
+				// grpc.WithDefaultServiceConfig(`{"loadBalancingPolicy": "pick_first"}`),
+			),
+		),
+	}
+
+	ctx := context.Background()
+	ctx, cancel := context.WithTimeout(ctx, 1*time.Second)
```
rekby:
```suggestion
	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
```

Why do you need a second?
Second is a lot of time, then you run a lot of tests. Can you decrease the timeout or use some signals for sync events instead of time (prefer)?

rekby (changes requested):

internal/balancer/balancer_test.go
kprokopenko: I couldn't find any events for synchronization, assuming that there might be retries with different logic and their own timeouts inside `discoveryConn`. If not now, then in the future. One second excludes false test failures in case the test is run on very slow workers. Moreover, one second is only lost if the test fails, and in that case, losing it is not critical. Upon successful completion of the test, the context is canceled via `cancel`.

internal/balancer/balancer_test.go
```diff
@@ -0,0 +1,98 @@
...
+	defer fakeListener.Close()
+
+	fakeServer := grpc.NewServer()
+	defer fakeServer.Stop()
+
+	go func() {
+		_ = fakeServer.Serve(fakeListener)
+	}()
+
+	var dialAttempt uint32
```
rekby: What about use atomic.Int32?

internal/balancer/balancer_test.go
```diff
@@ -0,0 +1,98 @@
...
+		address: "ydbmock:///mock",
+		driverConfig: config.New(
+			config.WithEndpoint("mock"),
+			config.WithGrpcOptions(
+				grpc.WithResolvers(&mockResolverBuilder{}),
+
+				grpc.WithContextDialer(
+					// The first dialing is never ended, while the subsequent ones work fine.
+					func(ctx context.Context, s string) (net.Conn, error) {
+						atomic.AddUint32(&dialAttempt, 1)
```
rekby: What about store returned result, for exactly one time make infinite dial?

for current code two parallel goroutines can increase the variable before atomic.Load read result. And no of goroutines will be infinite.

PR #1903, opened 2025-10-24
Correct sql.Null* support

1NepuNep1:
Supported sql.Null* types as params in `database/sql` query

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently, all `sql.Null*` values with `Valid: false` are parsed like `nil` without a type, cause of `Value()` method.

Issue Number: N/A

## What is the new behavior?

`Sql.Null*` value correctly parses into some valid type

-
-
-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 * Supported `sql.Null*` from `database/sql` as query params in `toValue` func
 
-## v3.117.3
+## v3.118.0
 * Added support for nullable `Date32`, `Datetime64`, `Timestamp64`, and `Interval64` types in the `optional` parameter builder
 * Added method `query.WithIssuesHandler` to get query issues
 
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -2,8 +2,8 @@ package version
 
 const (
 	Major = "3"
-	Minor = "117"
-	Patch = "2"
+	Minor = "118"
+	Patch = "0"
 
 	Package = "ydb-go-sdk"
 )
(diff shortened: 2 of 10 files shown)


internal/bind/params_test.go
```diff
@@ -998,3 +998,111 @@ func BenchmarkAsUUIDUsingReflect(b *testing.B) {
...
+		{
+			name:     "sql.Null[int64] valid",
+			src:      sql.Null[int64]{V: 42, Valid: true},
+			expected: value.OptionalValue(value.Int64Value(42)),
+		},
+		{
+			name:     "sql.Null[int64] invalid",
+			src:      sql.Null[int64]{V: 0, Valid: false},
+			expected: value.NullValue(types.Int64),
+		},
```
kprokopenko, line 1097: Предлагаю `sql.Null[T]` добить для всех поддерживаемых типов:

```
int64
float64
bool
[]byte
string
time.Time
```

internal/bind/params.go
```diff
@@ -47,6 +48,103 @@ func asUUID(v any) (value.Value, bool) {
...
+		
+		return value.NullValue(types.Bool), true
+	case sql.NullFloat64:
+		if x.Valid {
+			return value.OptionalValue(value.DoubleValue(x.Float64)), true
+		}
+
+		return value.NullValue(types.Double), true
+	case sql.NullInt16:
+		if x.Valid {
```
kprokopenko: Как будто бы: `if x.Valid { wrapWithNull(xxx) }` можно не повторять и вынести отдельно

PR #2029, opened 2026-02-19
feat(topic): log endpoint for topic writer session

kprokopenko:
**Main goal:** have the endpoint in the logs where the topic write session was established.

This makes it easier to understand on the server side what is happening: which node (node_id, address, location) is serving the writer session.

**Changes:**
- Topic writer init stream done/failed log events now include `node_id`, `address`, and `location` when the endpoint is available.
- gRPC client stream and rawtopic StreamWriter expose optional `Endpoint() endpoint.Endpoint` (not in public interfaces).
- Reconnector derives endpoint from stream and passes it into the writer config for tracing/logging.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/slo.yml
+++ b/.github/workflows/slo.yml
@@ -85,6 +85,15 @@ jobs:
             slo_workload_write_max_rps: 100
             run_extra_args: '-batch-size=10'
             create_extra_args: '-min-partitions-count 10'
+          - id: session_node_hints
+            name: session-node-hints
+            path: ./native/session_node_hints
+            label: native/session_node_hints
+            slo_workload_read_max_rps: 1000
+            slo_workload_write_max_rps: 1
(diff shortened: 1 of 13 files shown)


internal/topic/topicwriterinternal/writer_reconnector.go
```diff
@@ -610,13 +611,21 @@ func (w *WriterReconnector) waitFirstInitResponse(ctx context.Context) error {
 }
 
 func (w *WriterReconnector) createWriterStreamConfig(stream RawTopicWriterStream) SingleStreamWriterConfig {
```
rekby, line 612: Тут я бы включил `Endpoint` в интерфейс `RawTopicWriterStream`, тогда и использование метода будет проще найти и тут приведение интерфейсов не потребуется - достаточно будет endpoint просто взять.

PR #2096, opened 2026-04-23
fix(topic): do not create buffer for decoded message

abbit:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

New `bytes.Buffer` created for each decoded message.

## What is the new behavior?

- use `io.Reader` wrapper to put decoder back to pool after EOF
- fix putting `input` that implements `PublicResettableReader` to pool, when `CodecRaw` is used


internal/topic/topicreadercommon/decoders.go
```diff
@@ -17,86 +16,140 @@ type PublicResettableReader interface {
...
-func (d *MultiDecoder) AddDecoder(codec rawtopiccommon.Codec, createFunc PublicCreateDecoderFunc) {
-	d.m[codec] = createFunc
-	d.dp[codec] = newDecoderPool()
+func (w *resettableDecoderWrapper) Close() error {
+	if !w.closed {
+		w.closed = true
+		w.pool.Put(w.rd)
+		w.rd = nil
+		w.pool = nil
+	}
```
Danissimo-k, line 66: once.Do?

abbit: Go's documentation for `io.Reader` does not mention thread-safety, so there is no need to add additional overhead. Some boolean variable is needed anyway to check it in `Read` method.

kprokopenko (approved):

PR #2188, opened 2026-05-30
fix: do not ban connection on gRPC Canceled during query stream drain

### LINKED ISSUE #2186: streamWrapper must not ban connection on gRPC Canceled during ExecuteQuery drain
## Problem

Since v3.127.7 (#2040) `balancer.NewStream` wraps gRPC streams with `streamWrapper`, which calls `pool.Ban` on **any** `RecvMsg`/`SendMsg` error classified as a bad connection by `IsBadConn`.

`grpcCodes.Canceled` is treated as a bad connection. During normal query result drain (`result.Close`), the SDK may surface `Canceled` when tearing down the stream (e.g. `executeCancel` after reading all parts). That **bans the physical connection** even though the operation may have succeeded on the server.

Downstream RPCs on the same session/connection can then miss `ExecStats` in the first stream part (empty billing / flaky `CapacityUnits == 0` in docapi).

## Impact

- Docapi delete/put/update e2e flakiness after SDK v3.127.6 → v3.127.7 (query path, `ReturnConsumedCapacity`).
- Any client that drains commit streams after a prior stream ended with transport `Canceled`.

## How to reproduce

1. Use query service with multiple discovery endpoints (so the balancer can pick another connection after a ban).
2. Run **commit** query `select 1` with `StatsModeBasic` and `CommitTx` in a serializable RW transaction.
3. Configure the server (or a test double) so the **first** commit stream sends the first response part, then ends the stream with gRPC status **`Canceled`** on the next `Recv` (typical when the client drains/closes the stream after the first part).
4. Observe that the connection is **banned** (`conn` state / balancer pessimization).
5. Run the **same** commit again on the same session (or the next RPC on that session path).
6. Server sends `ExecStats` on the first stream part (normal commit billing), but the client may report **zero** consumed capacity / empty stats — e.g. docapi `ConsumedCapacity.CapacityUnits == 0` instead of the expected value.

On SDK **v3.127.6** (no `streamWrapper`, `pool.Ban` filtered by transport code) step 4 does not ban on `Canceled`, and step 6 is stable.

Bisect: regression appears between **v3.127.6** and **v3.127.7** (#2040).

## Expected behavior

- `grpc.Canceled` and `context.Canceled` during stream drain must **not** trigger connection pessimization (`pool.Ban` / `IsBadConn`).
- `ResourceExhausted`, `Unavailable`, and explicit operation-level overload signals should continue to ban as today.

## Proposed fix

- Exclude `grpcCodes.Canceled` from `IsBadConn` / `goodCodes`.
- Ignore `Canceled` causes in `pool.Ban`.

asmyasnikov:
## Pull request type

- [x] Bugfix

Fixes #2186

## Summary

Since **v3.127.7** (#2040), `balancer.NewStream` wraps every gRPC stream with `streamWrapper`. On any `RecvMsg`/`SendMsg` error classified as a “bad connection” by `IsBadConn`, the balancer calls `pool.Ban` and marks the physical connection as **Banned**.

**`grpc.Canceled` is classified as a bad connection.** That is incorrect for the normal client-side teardown of an `ExecuteQuery` stream: after the client reads the useful parts, `result.Close()` drains the stream and `executeCancel` cancels the gRPC context — the next `Recv` often returns `Canceled`. That is an expected outcome of drain, not a signal that the connection is unhealthy.

If the connection is banned, the **next** `ExecuteQuery` on the same session may still succeed on the server, but the client can observe **empty `ExecStats`** in `query.WithStatsMode` callbacks (zero read/write units) because stream delivery or endpoint selection is affected.

## Root cause

```mermaid
sequenceDiagram
    participant Client
    participant SDK as ydb-go-sdk
    participant Wrap as streamWrapper
    participant Pool as conn.Pool
    participant Srv as YDB

    Client->>SDK: ExecuteQuery (e.g. commit select 1)
    SDK->>Srv: gRPC stream
    Srv-->>SDK: response part 0
    Client->>SDK: result.Close(ctx)
    SDK->>Srv: Recv (drain)
    Srv-->>SDK: Canceled (stream teardown)
    SDK->>Wrap: RecvMsg error
    Wrap->>Pool: Ban(conn)
    Note over Pool: conn = Banned

    Client->>SDK: next ExecuteQuery on same session
    Note over Client,SDK: ExecStats may be missing in WithStatsMode callback
```

On **v3.127.6** there was no `streamWrapper`, and `pool.Ban` only accepted `ResourceExhausted` / `Unavailable` — `Canceled` during drain did not ban the connection.

## What changes

| File | Change |
|------|--------|
| `internal/balancer/errors.go` | Add `grpc.Canceled` to `goodCodes` → `IsBadConn` returns **false** for transport `Canceled`. |
| `internal/balancer/errors_test.go` | Expect `Canceled` → not a bad connection. |
| `internal/mock/commit_query.go`, `server.go` | In-process gRPC mock: reproducible commit streams for tests. |
| `internal/query/test/commit_consumed_capacity_test.go` | Regression via public `ydb.Open` + `query.Session` API. |
| `CHANGELOG.md` | `UNRELEASED` entry. |

**Not in scope:** `internal/conn/pool.go` is unchanged. `Pool.Ban` still bans unconditionally when invoked directly. This PR fixes the stream-drain path only, where `streamWrapper.onErr` (`balancer.go`) gates `pool.Ban` behind `IsBadConn`.

## Why the fix works

1. **`streamWrapper.onErr`** calls `pool.Ban` only when `IsBadConn(err)` is true. Excluding `grpc.Canceled` from bad transport codes stops drain-time cancellation from banning the connection on the `RecvMsg`/`SendMsg`/`CloseSend` wrapper path.
2. **Unchanged:** `ResourceExhausted`, `Unavailable`, and `BanOnOperationError` (e.g. `OVERLOADED`) still pessimise connections.
3. **Known limitation:** `conn.WithBanCallback` (`balancer.go`) invokes `pool.Ban` without an `IsBadConn` check. Call sites that pass a `Canceled` error through that callback would still ban the connection. No such call site is known for the query stream-drain scenario; adding a cancel filter to `Pool.Ban` itself is left for a follow-up if needed.

After the fix, the second commit on the same session receives `ExecStats` on the first stream part and the stats callback reports non-zero consumed units.

## Regression test

```bash
go test ./internal/query/test/ -run TestCommitConsumedUnitsAfterCanceledDrain -count=1 -v
```

**Scenario (public query API only):**

1. Local mock YDB with several discovery endpoints (balancer picks connections).
2. `query.Do` → `session.Begin` → transactional `Query("select 1", WithCommit(), WithStatsMode(Basic, ...))`.
3. **First** commit: mock sends part 0 (rows), then returns **`grpc.Canceled`** on the next server `Send` when the client drains via `Close`. `Query()` itself succeeds (first `Recv` in `newResult` gets part 0); the `Canceled` status is observed during drain, not as the `Query()` return error. The test intentionally ignores the first call's consumed units (zero — stats arrive only on the second commit) and relies on the mock's per-commit call counter (`commitQueryCalls == 1` after the first commit) to ensure the Canceled path was exercised before the second commit runs.
4. **Second** commit: mock sends **`ExecStats` on part 0**; test reads stats from the callback **before** `Close` (common commit pattern).
5. Assert consumed units == **2.0** — mock reports 1 delete row; write units use a ×2 multiplier (`1 row × 2 = 2.0`).

| Tree | Result |
|------|--------|
| `master` | **FAIL** — second commit often reports `capacity=0` |
| This branch | **PASS** — stable `capacity=2.0` |

```bash
go test ./internal/balancer/ -run TestIsBadConn -count=1 -v
```

`grpc.Canceled` must not be classified as a bad connection.

## Manual reproduction

1. SDK **v3.127.7+**, query client, multi-endpoint discovery.
2. Session + commit `select 1` with `WithStatsMode(StatsModeBasic)` and `WithCommit()`.
3. `Close` the result; stream ends with transport `Canceled` (normal).
4. On `master`, the connection becomes **Banned**.
5. Repeat the same commit; `WithStatsMode` may see empty stats despite server success.

After this PR, step 4 does not ban via the `streamWrapper`/`IsBadConn` path; step 5 gets consistent `ExecStats`.

See [#2186](https://github.com/ydb-platform/ydb-go-sdk/issues/2186).


internal/conn/pool.go
```diff
@@ -90,6 +91,12 @@ func (p *Pool) Ban(ctx context.Context, cc Conn, cause error) {
 		return
 	}
 
+	// gRPC/context Canceled on streaming drain must not ban the connection (docapi commit billing).
```
asmyasnikov:
```suggestion
	// Canceled (context or gRPC) is client- or server-initiated and does not indicate
	// a bad connection. Skip ban regardless of call path.
```

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## UNRELEASED
```
asmyasnikov:
```suggestion
```

internal/balancer/errors.go
```diff
@@ -34,6 +34,8 @@ var (
 		grpcCodes.ResourceExhausted,
 		grpcCodes.OutOfRange,
 		grpcCodes.OK,
+		// Client-side stream teardown (Close drain, executeCancel) must not ban the connection.
```
asmyasnikov:
```suggestion
		// client-side stream teardown (Close drain, executeCancel) must not ban the connection.
```

internal/balancer/errors.go
```diff
@@ -34,6 +34,8 @@ var (
 		grpcCodes.ResourceExhausted,
 		grpcCodes.OutOfRange,
 		grpcCodes.OK,
+		// client-side stream teardown (Close drain, executeCancel) must not ban the connection.
```
asmyasnikov:
```suggestion
```

internal/balancer/errors.go
```diff
@@ -60,7 +62,6 @@ func IsBadConn(ctx context.Context, err error, ignoreCodes ...grpcCodes.Code) bo
 	}
 
 	operationErrorCodes, _ := ctx.Value(ctxBanOnOperationError{}).(operationErrorCodesType)
```
asmyasnikov, line 63:
```suggestion
	operationErrorCodes, _ := ctx.Value(ctxBanOnOperationError{}).(operationErrorCodesType)

```

kprokopenko (approved):

PR #2189, opened 2026-05-30
fix: drain query stream after per-call context cancellation

### LINKED ISSUE #2187: Per-call context cancellation must not cancel executeCtx or poison query result stream
## Problem

`execute()` opens `ExecuteQuery` on a dedicated `executeCtx` (decoupled from the caller ctx via `xcontext.ValueOnly`) so the user can iterate the result with **different per-call contexts** (`NextResultSet(ctx)`, `Close(ctx)`).

In recent `streamResult` code, per-call ctx cancellation was forwarded to `executeCtx` via `streamCancel` (`context.AfterFunc` in `nextPart` / explicit call in `nextResultSet`). That:

1. Cancels the gRPC stream while the server may still send late parts (e.g. `ExecStats` in part 2).
2. Sets `lastErr` from `context.Canceled`, so subsequent `Close(freshCtx)` bails out before `Recv` — **late stream parts are never read**.

This breaks the contract that `Close(context.Background())` (or any live ctx) drains the stream after a failed `NextResultSet` on a cancelled/deadline ctx.

## Impact

- Clients using `query.WithStatsMode` may see empty stats when `ExecStats` arrive only in later stream parts and iteration was aborted with a cancelled per-call ctx.
- Docapi `ExecInQueryTransaction` commit: `transaction.Query("select 1", WithCommit())` + stats callback + `defer res.Close(ctx)` — if a read path uses a done ctx, billing can silently drop to 0.

## How to reproduce

Minimal scenario using **only the public query API** (no internal `nextPart`):

1. `session.Begin` with serializable read-write settings.
2. `tx.Query(ctx, "select 1", query.WithCommit(), query.WithStatsMode(query.StatsModeBasic, callback))`.
3. Server (or mock) returns the first stream part **without** `ExecStats`, and sends `ExecStats` only in the **second** part (common for commit responses).
4. Call `res.NextResultSet(cancelledCtx)` where `cancelledCtx` is already canceled — expect `context.Canceled`.
5. Call `res.Close(context.Background())` to drain the stream.

**Buggy behavior:** the stats callback is never invoked with real phase data; `Close` does not read the second part (stream already canceled / `lastErr` poisoned).

**Expected behavior:** step 4 fails with `context.Canceled`, but step 5 still drains the stream and the stats callback receives `ExecStats`.

Same pattern applies if step 4 is `NextRow` on a result set obtained earlier, or any other per-call read with a dead ctx followed by `Close` with a fresh ctx.

## Expected behavior

- `context.Canceled` / `DeadlineExceeded` on the **per-call** ctx passed to `NextResultSet` / row reads must:
  - return the ctx error to the caller;
  - **not** cancel `executeCtx`;
  - **not** set `lastErr` (so `Close` with another ctx can drain).
- `executeCancel` should run from normal `Close` completion (`withStreamResultOnClose`) or parent ctx cancellation during `execute()`, not from arbitrary per-call ctx on iteration.

## Proposed fix

- Remove forwarding of per-call ctx cancellation to `executeCtx` (`streamCancel` / `AfterFunc` in `nextPart` and `nextResultSet`).
- Do not poison `lastErr` on per-call context errors.
- Keep `withStreamResultOnClose(executeCancel)` for stream teardown after drain completes.

asmyasnikov:
## Pull request type

- [x] Bugfix

Fixes #2187

## Summary

`ExecuteQuery` runs on a dedicated **`executeCtx`**. Callers pass per-call contexts to `NextResultSet`, `NextRow`, and `Close`.

When a per-call context was cancelled or timed out during iteration, `streamResult` stored the error in **`lastErr`**. A later `Close(freshCtx)` returned that stale error **without draining** the stream, so late parts (for example `ExecStats` in a later `ExecuteQuery` part) were never read.

Typical failure: `WithStatsMode` callback stays empty after `NextResultSet(cancelledCtx)`, even though `Close(context.Background())` should drain the stream.

Per-call context cancellation still cancels the execute stream via `withCancelStream(executeCancel)` and `context.AfterFunc(perCallCtx, cancelStream)` in `nextPart` — this is intentional.

## What changes

| Area | Change |
|------|--------|
| `internal/query/result.go` | Do **not** set `lastErr` when the per-call context is already done (`ctx.Err() != nil`). `Close(freshCtx)` can drain late parts when the execute stream is still open. **`Close` returns `nil`** when the stream is already torn down (full iteration, per-call cancel, or repeated close) — same ergonomics as table results and `database/sql` auto-close on `NextResultSet` EOF. `Close` uses `context.AfterFunc(closeCtx, onClose)` so **`closeTimeout`** can unblock a stalled `Recv`. |
| `internal/query/execute_query.go` | Keep `withCancelStream(executeCancel)` and `withStreamResultOnClose(executeCancel)`. |
| Tests / mocks | Regressions for drain-after-cancel, idempotent `Close`, `closeTimeout`, public API commit+stats; `WithCommitStatsDelayed` mock behavior merged with master `#2188` helpers. |
| `CHANGELOG.md` | Two user-facing fix entries for `#2187`. |

## Why it works

1. Per-call ctx cancel/timeout still cancels the execute stream when registered (unchanged).
2. That error is **not** stored in `lastErr`, so it does not block a later drain on a live stream.
3. `Close(liveCtx)` runs the drain loop and delivers late stream parts (for example `ExecStats`).
4. `Close` on an already-closed stream is a no-op success (`onClose` still runs once via `sync.OnceFunc`).

## Regression tests

```bash
go test ./internal/query/test/ -run TestQueryResultCloseDrainsStatsAfterNextResultSetCanceledCtx -count=1 -v
go test ./internal/query/ -run 'CloseDrainsExecStats|CloseIsIdempotent|CloseTimeoutInterrupts|PerCallCtx' -count=1 -v
go test ./internal/xsql/test/ -run TestTwoStatementsTwoResultSets -count=1 -v
```

## When this matters

- `NextResultSet` / `NextRow` with a short-lived or cancelled ctx, then `Close` with another ctx.
- Commit queries with `WithStatsMode` when stats arrive in a later stream part.
- `database/sql` multi-result-set queries over query service.
- Session teardown paths that rely on `closeTimeout` during result drain.

See [#2187](https://github.com/ydb-platform/ydb-go-sdk/issues/2187).


internal/query/result.go
```diff
@@ -323,6 +300,9 @@ func (r *streamResult) Close(ctx context.Context) (finalErr error) {
 			if xerrors.Is(err, io.EOF) {
 				return nil
 			}
+			if ctxErr := ctx.Err(); ctxErr != nil {
+				return xerrors.WithStackTrace(ctxErr)
```
asmyasnikov:
```suggestion
				return ctxErr
```

internal/query/result.go
```diff
@@ -288,22 +254,33 @@ func nextPart(stream Ydb_Query_V1.QueryService_ExecuteQueryClient) (
...
 
 func (r *streamResult) Close(ctx context.Context) (finalErr error) {
-	if !r.closed.CompareAndSwap(false, true) {
-		return nil
+	if r.stream != nil {
+		if streamCtx := r.stream.Context(); streamCtx != nil {
+			if err := streamCtx.Err(); err != nil {
+				return nil
+			}
+		}
```
asmyasnikov:
```suggestion
		if r.stream.Context().Err() != nil {
			return nil
		}
```

internal/xsql/test/bench_test.go
```diff
@@ -79,9 +79,9 @@ func benchmarkDatabaseSQLSelect42(b *testing.B, nativeDriver *ydb.Driver, useQue
 // cpu: Apple M3 Pro
 // go test -bench=. -benchtime=10s .
 //
-// BenchmarkDatabaseSQL/over/QueryService-12		10475 ns/op     22580 B/op      371 allocs/op
-// BenchmarkDatabaseSQL/over/TableService-12		9493 ns/op      19131 B/op      306 allocs/op
-// Diff (query/table*100-100)						10%				18%				21%
+// BenchmarkDatabaseSQL/over/QueryService-12		10318 ns/op     21436 B/op      354 allocs/op
+// BenchmarkDatabaseSQL/over/TableService-12		9694 ns/op      19281 B/op      310 allocs/op
+// Diff (query/table*100-100)						        6.4%			      11%				      14%
```
asmyasnikov:
```suggestion
// BenchmarkDatabaseSQL/over/QueryService-12	10318 ns/op   21436 B/op      354 allocs/op
// BenchmarkDatabaseSQL/over/TableService-12	9694 ns/op    19281 B/op      310 allocs/op
// Diff (query/table*100-100)					6.4%		  11%			  14%
```

internal/table/test/bench_test.go
```diff
@@ -65,9 +65,9 @@ func warmUp(ctx context.Context, t testing.TB, driver *ydb.Driver) {
 
 // BenchmarkTable measures table.Client (driver.Table()) against the in-process xsql mock: Session.Execute("SELECT 42").
 //
-// BenchmarkTable/over_query-client-12      10264 ns/op    20903 B/op   352 allocs/op
-// BenchmarkTable/original_table-client-12  11132 ns/op    19877 B/op   335 allocs/op
-// Diff (query/table*100-100)				-7.8%		   5%			5%
+// BenchmarkTable/over_query-client-12      11929 ns/op    21062 B/op   358 allocs/op
+// BenchmarkTable/original_table-client-12  10962 ns/op    20019 B/op   339 allocs/op
+// Diff (query/table*100-100)				        8.8%			     5.2%		      5.6%
```
asmyasnikov:
```suggestion
// BenchmarkTable/over_query-client-12      11929 ns/op    21062 B/op   358 allocs/op
// BenchmarkTable/original_table-client-12  10962 ns/op    20019 B/op   339 allocs/op
// Diff (query/table*100-100)				8.8%		   5.2%		    5.6%
```

internal/query/result.go
```diff
@@ -288,22 +270,23 @@ func nextPart(stream Ydb_Query_V1.QueryService_ExecuteQueryClient) (
...
-		return nil
-	}
+	defer r.onClose()
 
-	defer func() {
-		for _, f := range r.shutdownHooks {
-			f()
+	if r.stream != nil {
+		if r.stream.Context().Err() != nil {
+			return nil
```
asmyasnikov:
```suggestion
		if ctxErr := r.stream.Context().Err(); ctxErr != nil {
			return ctxErr
```

internal/mock/commit_query.go
```diff
@@ -69,6 +76,19 @@ func (m *querySrv) executeCommitQuery(
...
+			Status:         Ydb.StatusIds_SUCCESS,
+			ResultSetIndex: 0,
+			ResultSet:      selectOneResultSet(),
+		}); err != nil {
+			return err
+		}
+
+		return stream.Send(&Ydb_Query.ExecuteQueryResponsePart{
+			Status:    Ydb.StatusIds_SUCCESS,
+			ExecStats: docapiCommitExecStats(),
```
kprokopenko: Как будто бы подсвечиваем нашу внутрянку про `docapi`. Лучше что-нибудь нейтральное выбрать в качестве имени.

kprokopenko (approved):

internal/mock/commit_query.go
asmyasnikov: fixed

PR #2201, opened 2026-06-17
Fix consumer alters silently resetting supported codecs

nikolaymatrosov:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

`AlterConsumer.ToProto()` populates `set_supported_codecs` unconditionally, and
`SupportedCodecs.ToProto()` never returns nil — it returns a non-nil proto with an
empty codec list even for a nil slice. As a result, any unrelated consumer alter
via the public `Topic().Alter()` API — `AlterConsumerWithImportant`,
`AlterConsumerWithReadFrom`, `AlterConsumerWithAttributes`,
`AlterConsumerWithAvailabilityPeriod` — sends `set_supported_codecs: []`
("allow any") for that consumer, silently wiping its existing codec restriction
server-side.

Issue Number: N/A

## What is the new behavior?

- `set_supported_codecs` is sent only when `AlterConsumerWithSupportedCodecs` is
  used; unrelated consumer alters no longer touch the codec restriction.
- Guards the field with a presence flag (`SetSupportedCodecs bool` +
  `SetSupportedCodecsValue`), mirroring the existing topic-level path.
- Adds regression tests at both the `ToProto` level and the public-options level,
  plus a `CHANGELOG.md` entry.

## Other information

The explicit "reset to allow-any" path is preserved: calling
`AlterConsumerWithSupportedCodecs(name, nil)` still sends `set_supported_codecs: []`.
The topic-level alter path was already correct (it guards with the same flag); this
change brings the consumer-level path in line with it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,6 @@
 
 ## v3.139.7
 * Fixed YSON scanning in `TableService` to support both underlying `TextValue` and `BytesValue` wire representations
-
 * Fixed inverted success/error handling in the `ExampleWriter_Write` doc example for `topicwriter`, which printed `OK` on failure and aborted on success
 * Fixed nil pointer dereference panic in `topicsugar.ProtobufIterator` on the first received message by allocating a concrete protobuf message before unmarshaling
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+* Fixed `Topic().Alter()` consumer alters (`AlterConsumerWithImportant`, `AlterConsumerWithReadFrom`, `AlterConsumerWithAttributes`, `AlterConsumerWithAvailabilityPeriod`) silently resetting the consumer's supported-codecs restriction: `set_supported_codecs` is now sent only when `AlterConsumerWithSupportedCodecs` is used
+
 ## v3.140.0
 * Added `topicoptions.WithWriterDirectWrite(bool)` and `topicoptions.WithMultiWriterDirectWrite(bool)` options to send topic writes to the node that hosts the target partition, bypassing the topic proxy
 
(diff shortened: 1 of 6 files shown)


internal/grpcwrapper/rawtopic/alter_topic.go
```diff
@@ -85,19 +85,23 @@ type AlterConsumer struct {
 	Name                    string
 	SetImportant            rawoptional.Bool
 	SetReadFrom             rawoptional.Time
-	SetSupportedCodecs      rawtopiccommon.SupportedCodecs
+	SetSupportedCodecs      bool
+	SetSupportedCodecsValue rawtopiccommon.SupportedCodecs
```
kprokopenko:
```suggestion
	SetSupportedCodecs      rawtopiccommon.SupportedCodecs
	NeedSetSupportedCodecs bool
```

Уменьшит diff и поддержит текущие имена полей вида `SetXXX`.

kprokopenko (changes requested):
Я бы еще добавил регрессионный интеграционный тест, который подсвечивает решаемую проблему.

kprokopenko (approved):

CHANGELOG.md
```diff
@@ -6,6 +8,7 @@
 
 ## v3.139.7
 * Fixed YSON scanning in `TableService` to support both underlying `TextValue` and `BytesValue` wire representations
+
 * Fixed inverted success/error handling in the `ExampleWriter_Write` doc example for `topicwriter`, which printed `OK` on failure and aborted on success
```
kprokopenko, line 11:
```suggestion
* Fixed YSON scanning in `TableService` to support both underlying `TextValue` and `BytesValue` wire representations
* Fixed inverted success/error handling in the `ExampleWriter_Write` doc example for `topicwriter`, which printed `OK` on failure and aborted on success
```

